PR #11, opened 2019-04-04
added docs for trait Service, trait Transform

Dowwie:


actix-service/src/transform.rs
```diff
@@ -33,7 +40,9 @@ pub trait Transform<S> {
     /// The future response value.
     type Future: Future<Item = Self::Transform, Error = Self::InitError>;
 
-    /// Create and return a new service value asynchronously.
+    /// Creates and returns a new Service component, asynchronously.  `new_transform`
+    /// is called only once during the lifetime of a server -- as the server
```
fafhrd91: this is not right. you can create any number of services. once per server lifetime is specific for http server impl.

actix-service/src/lib.rs
```diff
@@ -53,6 +53,9 @@ pub use self::then::{Then, ThenNewService};
 pub use self::transform::{apply_transform, IntoTransform, Transform};
 
 /// An asynchronous function from `Request` to a `Response`.
+/// The entire request/response lifecycle can be represented as a pipeline of
+/// Services.  The linkage between Services is represented in terms of ownership,
```
fafhrd91: i am not sure this belongs to Service docs. Service trat describers function call and that is it.

PR #15, opened 2019-04-29
Add compatibility layer for tower-service

neoeinstein:
It seems that `actix-service` is largely based on `tower-service`. There are a lot of great middlewares already being built over in the [`tower-rs/tower`](https://github.com/tower-rs/tower) repository, and it would be nice to be able to easily use those middlewares here. In order to do that, we need some sort of compatibility layer.

I see two ways of making this easy. The first is to add a feature flag to `actix-service` which would bring in the `tower-service` dependency and provide the compatibility layer. This would make the dependency optional, so an end user would need to specificy the feature to see the compatibility layer. The second would be to add a new crate, something like `actix-tower`, which would then depend on `actix-service` and `tower-service`, and provide the compatibility traits that way.

I decided to go the first route for this pass, but could easily be convinced to go the second route.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,6 +23,7 @@ members = [
   "actix-server-config",
   "actix-test-server",
   "actix-threadpool",
+  "actix-tower",
   "actix-utils",
   "router",
 ]
--- a/actix-tower/CHANGES.md
+++ b/actix-tower/CHANGES.md
@@ -1,0 +1,2 @@
+# Changes
+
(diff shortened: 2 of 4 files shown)


fafhrd91:
Let’s move it separate crate. I don’t want to add dependencies to actix-service

neoeinstein:
Understood. I've just updated the commit to move it out into its own crate. I've included the `CHANGES.md` file, but have not populated it.

fafhrd91:
@neoeinstein i've decided to drop tower integration, if you'd like to maintain it please clone it to different repo.

PR #20, opened 2019-05-18
Add ServerCommand::Restart and Command::Restart

joshleeb:
This PR creates a new `ServerCommand` and corresponding `Command` to restart server workers.

Related to actix/actix-web#280

### DIFF
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -16,6 +16,7 @@ use super::Token;
 pub(crate) enum Command {
     Pause,
     Resume,
+    Restart,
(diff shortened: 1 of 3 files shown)


joshleeb:
I'm not sure if this PR is on the right track as I'm not all that familiar with `mio::Poll`. Happy to fix this up in anyway or just close it if I'm not at all close to what needs to be done.

fafhrd91:
I think restart should actually restart worker threads, keeping sockets without change

fafhrd91:
closing as inactive

PR #22, opened 2019-05-28
Allow Actix to be started on an external CurrentThread runtime

GeorgeHahn:
This PR includes the minimum set of changes to enable running Actix on an external runtime.

Example usage:

```
let rt = tokio_current_thread::CurrentThread.handle();
let system = actix::System::new_async("async system", rt.to_owned());
rt.spawn(system.run_nonblocking().map_err(|e| error!("{}", e)))
  .map_err(|e| error!("Could not start actix async: {}", e))
```

### DIFF
--- a/actix-rt/src/builder.rs
+++ b/actix-rt/src/builder.rs
@@ -73,7 +73,7 @@ impl Builder {
     /// Create new System that can run asynchronously.
     ///
     /// This method panics if it cannot start the system arbiter
-    pub fn build_async(self, executor: Handle) -> AsyncSystemRunner {
+    pub(crate) fn build_async(self, executor: Handle) -> AsyncSystemRunner {
         self.create_async_runtime(executor)
     }
 
@@ -87,8 +87,7 @@ impl Builder {
         self.create_runtime(f).run()
     }
 
-    fn create_async_runtime(self, executor: Handle) -> AsyncSystemRunner
-    {
+    fn create_async_runtime(self, executor: Handle) -> AsyncSystemRunner {
         let (stop_tx, stop) = channel();
         let (sys_sender, sys_receiver) = unbounded();
 
@@ -152,7 +151,7 @@ impl Builder {
 }
 
 #[derive(Debug)]
-pub struct AsyncSystemRunner {
+pub(crate) struct AsyncSystemRunner {
     stop: Receiver<i32>,
     system: System,
 }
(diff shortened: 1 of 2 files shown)


fafhrd91:
i think `AsyncSystemRunner` should be internal impl. System::new_async should just return impl Future

actix-rt/src/builder.rs
```diff
@@ -127,6 +151,48 @@ impl Builder {
...
+}
+
+impl AsyncSystemRunner {
+    /// This function will start event loop and returns a future that
+    /// resolves once the `System::stop()` function is called.
+    pub fn run_nonblocking(self) -> Box<Future<Item = (), Error = io::Error> + Send + 'static> {
+        let AsyncSystemRunner { stop, .. } = self;
+
+        // run loop
+        Box::new(future::ok(())
```
fafhrd91: i think you can collapse most of .and_then and put everything into `future::lazy()`

actix-rt/src/system.rs
```diff
@@ -55,6 +56,14 @@ impl System {
         Self::builder().name(name).build()
     }
 
+    #[allow(clippy::new_ret_no_self)]
+    /// Create new system using provided CurrentThread Handle.
+    ///
+    /// This method panics if it can not spawn system arbiter
+    pub fn new_async<T: Into<String>>(name: T, executor: Handle) -> AsyncSystemRunner {
```
fafhrd91: i think this should be something like `run_in_executor`

GeorgeHahn:
Thanks for the review! I've made the changes you requested & tested them out locally.

fafhrd91:
thanks!

GeorgeHahn:
@mmacedoeu Nothing in-tree, but here's an updated version of the example at the top of this PR:

```
let rt = tokio_current_thread::CurrentThread.handle();
let system = actix::System::run_in_executor("async system");
rt.spawn(system.map_err(|e| error!("Error from actix system: {}", e)));
```

Actix will be fully functional after this code runs.

mmacedoeu:
I think when you pass the handle to run_in_executor it already spawn system.
I am trying to send a message using send from an Addr returned after a SyncArbiter::start(...) then spawn send Request as usual but Message is not processed. Could you elaborate an example where a message is spawned to an Addr. I think something is missing here.

GeorgeHahn:
Passing the handle into `run_in_executor` allows actix to spawn the system arbiter but does not run the actix system. The returned future must be spawned on a tokio runtime.

Here's a full example:
`Cargo.toml`
```toml
[package]
name = "actix-test"
version = "0.1.0"
authors = ["George Hahn <george.hahn.vhs@gmail.com>"]
edition = "2018"

[dependencies]
actix = "0.8"
tokio = "0.1"
tokio-current-thread = "0.1"
log = "*"
futures = "0.1"
```

`main.rs`
```rust
use actix::prelude::*;
use futures::Future;
use log::error;
use tokio_current_thread::CurrentThread;

struct Ping(usize);

impl Message for Ping {
    type Result = usize;
}

struct MyActor {
    count: usize,
}

impl Actor for MyActor {
    type Context = Context<Self>;
}

impl Handler<Ping> for MyActor {
    type Result = usize;

    fn handle(&mut self, msg: Ping, _: &mut Context<Self>) -> Self::Result {
        self.count += msg.0;
        self.count
    }
}

fn main() {
    let mut rt = CurrentThread::new();
    let handle = rt.handle();
    let system = actix::System::run_in_executor("async system", rt.handle());
    handle
        .spawn(system.map_err(|e| error!("Error from actix system: {}", e)))
        .expect("start actix");

    let addr = MyActor { count: 10 }.start();
    let res = addr.send(Ping(10));

    handle
        .spawn(
            res.map(|res| {
                println!("RESULT: {}", res == 20);
            })
            .map_err(|_| ()),
        )
        .unwrap();

    rt.run().unwrap();
}
```

PR #30, opened 2019-07-16
Expose Connect addrs

jan-auer:
Exposes the internal addresses of the `Connect` request to allow composing `Resolvers` on top of the existing one. There are now two new methods: `Connect::addrs` and `Connect::take_addrs`.

My use case is that I need to prevent the `TcpConnector` from opening a connection to certain IP ranges that have been resolved by the `Resolver`. This can be neatly achieved by inserting a service in between:

```rust
fn filter_ip_addrs<T: Address>(mut connect: Connect<T>) -> Result<Connect<T>, ConnectError> {
    let mut addrs = connect
        .take_addrs()
        .filter(|addr| !RESERVED_NETWORKS.iter().any(|net| net.contains(addr.ip())))
        .peekable();

    if addrs.peek().is_none() {
        return Err(ConnectError::NoRecords);
    }

    Ok(connect.set_addrs(addrs))
}

awc::Client::build().connector(
    awc::Connector::new().connector(
        Resolver::default()
            .and_then(filter_ip_addrs)
            .and_then(TcpConnector::new())
    ).finish()
).finish();
```

cc @mitsuhiko

### DIFF
--- a/actix-connect/src/connect.rs
+++ b/actix-connect/src/connect.rs
@@ -1,5 +1,6 @@
-use std::collections::VecDeque;
+use std::collections::{vec_deque, VecDeque};
 use std::fmt;
+use std::iter::{FromIterator, FusedIterator};
 use std::net::SocketAddr;
 
 use either::Either;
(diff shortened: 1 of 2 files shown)


fafhrd91:
actix-connect 0.2.1 is released

PR #42, opened 2019-09-09
Update rustls, tokio-rustls and webpki across the board

jeromegn:


jeromegn:
I'm not sure how I can make this pass given the way dependencies are declared.

jeromegn:
I think this is the kind of PR that needs to be merged in and the manually publish the individual crates and update their dependencies? Not sure how else this can work, short of pointing dependencies within actix-net using paths instead of versions from crates.io.

If there's anything I can do, please let me know!

guoli-lyu:
Update rustls to 0.16 entails major version bump in ring. And linking to ring-asm with different versions of ring are no allowed.

So the solution is fairly clear. All dependencies must update to ring 0.16, which is not an trivial job for breaking API in ring.

jeromegn:
> So the solution is fairly clear. All dependencies must update to ring 0.16, which is not a trivial job because of breaking API changes in ring.

@guoli-lyu It doesn't appear too bad. Looks like the hkdf and hmac stuff has a new API, but in actix-web there's barely any uses of ring directly.

There's nothing preventing the change in actix-net though, no use of ring directly. actix-web will need to upgrade if it wants to start using the latest actix-net sub-crates, but that feels outside the scope of this project? I'm not sure, I'm new here :)

svenstaro:
Can you rebase this?

svenstaro:
@fafhrd91 if you're generally happy with these changes, I'd like to get the whole actix stack onto rustls 0.16 soon.

fafhrd91:
change doesnt matter, new version of ring is hard breaking change.

JohnTitor:
Even if rebasing it, we still have a problem with dependencies, I think.

JohnTitor:
Wait, it includes unrelated commits?

jeromegn:
@JohnTitor Hmm, somehow my rebase got weird. Fixing.

svenstaro:
It would seem that this needs a bump of the minimum rust version.

JohnTitor:
It seems we need 1.37.0 or later?

JohnTitor:
I'd like to hear opinion from @fafhrd91

jeromegn:
`ring` bump is indeed a breaking change, but only for `actix-web`.

I've been planning on switching our non-actix codebase to using actix, but I need actix-net and I need rustls. I'm fine with using this fork for a while, I assume rustls will get updated at some point in here.

Either way, happy to take this slow and continue making requested changes.

jeromegn:
Updated. I wasn't sure how to name an unreleased entry in the changelogs though. I did what some other repos do.

svenstaro:
Well, let's go then? @JohnTitor

JohnTitor:
I think this can go but I'm not sure @fafhrd91 approves bumping up the minimum supported version.

svenstaro:
Isn't this necessarily going to happen sooner or later?

fafhrd91:
lets merge but we must bump minor version for all three crates

JohnTitor:
@fafhrd91 Just confirmation: I'll make new releases, minor version means `0.x.0`?

fafhrd91:
@JohnTitor yes

JohnTitor:
Ugh, unfortunately, I cannot push commit to master because of the branch protection.

fafhrd91:
create pr, i'll approve

PR #43, opened 2019-09-12
Feature uds: Add listen_uds to ServerBuilder

karlri:
Allows directly passing an Unix Listener instead of a path. Useful
for example when running as a daemon under systemd with the systemd
crate.


karlri:
Force-pushed a line to CHANGES.md

PR #47, opened 2019-09-14
Migrate actix-rt to std::future

semtexzv:
This has several issues, the current implementation pulls whole of tokio 0.2, which
should be replaced by tokio-runtime crate, once its ready. It also adds
a single unsafe block which performs pinning of boxed futures for spawning
them on an executor


actix-rt/src/arbiter.rs
```diff
@@ -321,14 +335,13 @@ impl SystemArbiter {
...
-    type Error = ();
+    type Output = ();
 
-    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         loop {
-            match self.commands.poll() {
-                Ok(Async::Ready(None)) | Err(_) => return Ok(Async::Ready(())),
-                Ok(Async::Ready(Some(cmd))) => match cmd {
+            match unsafe { self.as_mut().map_unchecked_mut(|p| &mut p.commands) }.poll_next(cx) {
```
Fedott:
```suggestion
            match Pin::new(&mut self.rx).poll_next(cx) {
```
You do not need to use unsafe

semtexzv: Thanks, fixed it.

PR #49, opened 2019-09-19
Migrate actix-service to std::future

semtexzv:
Whew ...

Big picture:
Using `Pin<&mut Self>` in `poll_ready` since it usually polls futures, not in call, which we might wanna change. There is an issue with the IntoFuture trait not being provided by futures anymore, so I wrote a simple one, and am using it throughout the codebase. This is just a first iteration, and will require some work, also BoxedService impl currently does not work, since it polls a future in `call` method


fafhrd91:
Sorry, for slow response, I don’t have time at the moment. I’ll be able to do proper review in 2 weeks

cdbattags:
Updated std-future to the current work so far here and let's close this potentially so we aren't confused?

cdbattags:
Closing and copying discussion to #57.

PR #50, opened 2019-09-30
Migrate actix-server to std::future

semtexzv:
This is less-complete than other parts, Since I don't have that much time lately, but I still wanted to submit it for review, and maybe  for somebody else to help.

Codebase is compilling, some tests are passing. This depends on #49


fafhrd91:
good work!

PR #57, opened 2019-11-07
Migrate actix-service to std::future

cdbattags:
**Originally from @semtexzv.**

> The big picture starting with actix-service:
> 
> Using Pin<&mut Self> in poll_ready since it usually polls futures, not in call, which we might wanna change. There is an issue with the IntoFuture trait not being provided by futures anymore, so I wrote a simple one, and am using it throughout the codebase. This is just a first iteration, and will require some work, also BoxedService impl currently does not work, since it polls a future in call method.
> 
> Decisions that we need to make:
> 
> 1. The old future trait had Item, and Error associated types. This meant it was similar to the Result type, and thus, every function that returned future was fallible. Now, the future has only one associated type, that denotes infallible future output.
> 
> Several places return a future with a result, that is not used anywhere, or its error type is (). Which of these occurences should be replaced with Future<Output=Item>, and which with Future<Output=Result<Item,Error>>?
> 
> 2. Usage of the Pin<&mut Self> in places similar to future, this means the Service::poll_ready and ActorFuture::poll in actix.
> 
> There is a reason for using these Pins in the poll method, should we upgrade our definitions of
> traits similar to Future to also utilize them ?. Also, maybe not, since these were primarily introduced to support await in the form of generators. Need more info from qualified people.
> 
> 3. Usage and form of macros / functions to ease pinning and unpinning.
> 
> I am currently working on actix-service, and it is a massive chore to always create unsafe block in
> order to create or destrucure a Pin<&mut T>. This can be sometimes solved by pin projections from the pin-utils crate, but it does not allow splitting borrows ( Creating multiple pins to multiple fields of a struct), which is a massive pain.


cdbattags:
Looks like only current build errors are:

image

cdbattags:
Big picture starting with `actix-service`:

Using Pin<&mut Self> in poll_ready since it usually polls futures, not in call, which we might wanna change. There is an issue with the IntoFuture trait not being provided by futures anymore, so I wrote a simple one, and am using it throughout the codebase. This is just a first iteration, and will require some work, also BoxedService impl currently does not work, since it polls a future in call method.

cdbattags:
Quick errors to get rid of by replacing `Poll` with lowercase `poll` in new futures-rs `0.3.1`.

semtexzv:
I'd rather import `Poll` from std::task. In general try to use types from std if possible. Even though this is just a re-export.

cdbattags:
@semtexzv, definitely agree to use `std` wherever we can

I'll go through and refactor those bits ideally this weekend

fafhrd91:
Before doing any changes, you should clearly understand what it is and how it works

cdbattags:
For the future associated type, I say the fastest refactor would be to always default to using `Future<Output=Result<Item,Error>>` and still tossing in `()` for error in the first pass when we don't need it.

cdbattags:
> Before doing any changes, you should clearly understand what it is and how it works

Thank you for the warning? I'm here to help, my sir.

fafhrd91:
I don’t want to offend anybody, but i think that understanding of the system is super important. Whole actix stack is built on top of actix-service. We must be super careful.

cdbattags:
> I don’t want to offend anybody, but i think that understanding of the system is super important. Whole actix stack is built on top of actix-service. We must be super careful.

Understood. So let's start some documentation around your warnings and your thoughts so far? Github wiki or Google Doc or what would you prefer?

cdbattags:
Maybe best to go ahead and set up some tests to attempt to maintain some kind of concept of backwards compatibility? Do we want to completely leave 0.1 futures behind and do we want to only support tokio 2.0 from here on?

synaestheory:
I'm sorry if this is naive; Since the futures crate seems to have dropped the `IntoFuture` trait, they've introduced a struct: [futures::future::IntoFuture](https://docs.rs/futures/0.3.1/futures/future/struct.IntoFuture.html). Would it be better to use that than to stick an `IntoFuture` trait?

fafhrd91:
i'd like to do all work std-future branch

PR #46, opened 2019-09-14
Migrate threadpool to std::future

semtexzv:
Contains code to migrate actix-threadpool to std::future, overall progress tracked in #45.

Issue #45, opened 2019-09-14
Tracking issue for std::future migration


semtexzv:
This issue tracks the migration.

Crates migrated:

- [x] - actix-threadpool #46 
- [x] - actix-rt  #47 
- [x] - actix-codec #48 
- [x]  - actix-service (discussion needed) #57  
- [x]  - actix-server-config (Needs bugfixing)
- [x]  - actix-server (Needs bugfixing)
- [x]  - actix-utils
- [x]  - actix-connect
- [x]  - actix-ioframe

note: these are in rough dependency order

Decisions:

1. The old future trait had `Item`, and `Error` associated types. This meant it was similar to the Result type, and thus, every function that returned future was fallible. Now, the future has only one associated type, that denotes infallible future output.

    Several places return a future with a result, that is not used anywhere, or its error type is `()`. Which of these occurences should be replaced with `Future<Output=Item>`, and which with `Future<Output=Result<Item,Error>>`?

2.  Usage of the `Pin<&mut Self>` in places similar to future, this means the `Service::poll_ready` and `ActorFuture::poll` in actix. 

    There is a reason for using these Pins in the poll method, should we upgrade our definitions of 
traits similar to `Future` to also utilize them ?. Also, maybe not, since these were primarily introduced to support await in the form of generators. Need more info from qualified people.

3. Usage and form of macros / functions to ease pinning and unpinning. 
    
    I am currently working on actix-service, and it is a massive chore to always create unsafe block in 
order to create or destrucure a `Pin<&mut T>`. This can be sometimes solved by pin projections from the `pin-utils` crate, but it does not allow splitting borrows ( Creating multiple pins to multiple fields of a struct), which is a massive pain.

fafhrd91:
Lets crate Pr per crate. Do you want to do dev actix-net repo?

fafhrd91:
We can also skip actix-tower, I’d like to remove it

Jonathas-Conceicao:
I have played a little around moving some of the crates here to 0.3, I don't have anything final but I would also like to help.

About your points here's my opinion:
>1. Which of these occurences should be replaced with Future<Output=Item>, and which with Future<Output=Result<Item,Error>>?

Values that return the same time should not use a `Result`. But for values that does need a `Result` return type I think it's worth using the [TryFuture](https://rust-lang-nursery.github.io/futures-api-docs/0.3.0-alpha.18/futures/future/trait.TryFuture.html) trait.

[...]

PR #59, opened 2019-11-10
Migrate actix-threadpool to std::future

tyranron:
Part of #45

This PR supersedes and continues #46 by extending the work of @semtexzv.  
I cannot push to that branch that's why I've created a separate PR.

## What's changed?
- `CpuFuture` now doesn't deal with user errors, just maps the result of user's future transparently.
- Unnecessary `unsafe` is removed when dealing with Pin inside `Future`'s impl for `CpuFuture` (comparing to `std-future` branch).
- `log::error!` is replaced with `log::warn!` as, in fact, creating thread pool is not fallible at all, and error of parsing threads from env var is just backed by default value. No falling - no error, but warning is.

## Checklist
- [x]  `rustfmt` applied
- [x] crate's version bumped
- [x] changelog pre-filled


tyranron:
Well... the `Poll` import should be fixed across the all crates. I'd better do it as a separate PR after merging this one.

PR #60, opened 2019-11-12
Allow to join the Arbiter

tsionyx:
I had an [issue](https://github.com/actix/actix/issues/296) with actors running into non-default Arbiter, so this is a way to wait for the Arbiter to finish by joining its underlying thread.


tsionyx:
Added entry to CHANGES.md. 

And about `std-future`: should I create another PR?

fafhrd91:
thanks!

let's create separate PR for `std-future`

PR #61, opened 2019-11-12
Migrate actix-server::ssl::nativetls to std futures

tyranron:
### actix-server-config

- Remove `"uds"` feature (`tokio` provides it along with its default features).
- Rename features: `"ssl"` -> `"openssl"`, `"rust-tls"` -> `"rustls"` (for being more clear).
- Unmake features being enabled by default (as `master` has).
- Remove unused dependencies.

### actix-server

- Migrate `ssl::nativetls` module to `std::future` via using [`tokio-tls`](https://docs.rs/tokio-tls/0.3.0-alpha.6/tokio_tls) crate. The previous implementation cannot be transported "as is" due to necessity of passing `Waker` explicitly (previously it was taken from TLS under-the-hood).
- Rename features: `"tls"` -> `"nativetls"`, `"ssl"` -> `"openssl"`, `"rust-tls"` -> `"rustls"` (for consistency with `actix-server-config` and modules names).

### DIFF
--- a/actix-server/src/ssl/nativetls.rs
+++ b/actix-server/src/ssl/nativetls.rs
@@ -1,16 +1,18 @@
+use std::convert::Infallible;
+use std::future::Future;
 use std::io;
 use std::marker::PhantomData;
+use std::pin::Pin;
+use std::task::{Context, Poll};
 
-use actix_service::{NewService, Service};
-use futures::{future::ok, future::FutureResult, Async, Future, Poll};
-use native_tls::{self, Error, HandshakeError, TlsAcceptor};
+use actix_service::{Service, ServiceFactory};
+use futures::future;
+use native_tls::{Error, HandshakeError, TlsAcceptor, TlsStream};
 use tokio_io::{AsyncRead, AsyncWrite};
 
 use crate::counter::{Counter, CounterGuard};
 use crate::ssl::MAX_CONN_COUNTER;
 use crate::{Io, Protocol, ServerConfig};
-use std::pin::Pin;
-use std::task::Context;
 
 /// Support `SSL` connections via native-tls package
 ///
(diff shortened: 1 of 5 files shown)


tyranron:
@fafhrd91 I'll finish other `ssl` modules once this PR will be merged, because there are some questions to be discussed, which I'll describe below in shortly.

actix-server/src/ssl/nativetls.rs
```diff
@@ -68,130 +80,46 @@ pub struct NativeTlsAcceptorService<T, P> {
...
-            Ok(_) => (),
-            Err(ref e) if e.kind() == io::ErrorKind::WouldBlock => (),
-            Err(e) => return Err(e),
-        }
-        self.inner.get_mut().shutdown()
+    fn call(&mut self, req: Self::Request) -> Self::Future {
+        let guard = self.conns.get();
+        let this = self.clone();
+        let (io, params, proto) = req.into_parts();
+        async move { this.acceptor.accept(io).await }
```
tyranron: @fafhrd91 here is the place I don't really like, because it imposes unnecessary `'static` bounds on type parameters and requires additional cloning. We're made to do so, however, because of this method signature:
```rust
fn call(&mut self, req: Self::Request) -> Self::Future
```

In fact, if Rust would have supported `async fn` in traits, it would be
```rust
async fn call(&mut self, req: Self::Request) -> Result<Self::Response, Self::Error>
```
and in such a way would require no `'static`/`.clone()`ing at all.

Seems being same, but the point is in lifetimes elision rules. `async fn` output captures lifetimes of all its parameters lifetimes into the output, while our function does not capture any which results in "outlive" error.

Would it be acceptable to bikeshed lifetimes of `Service::call()` method to make it more usable along with `async fn`s?

fafhrd91: we can bikeshed, but i dont have any idea about async call.

fafhrd91: i dont think you can remove extra clone. borrow checker is still there, so you would need to do clone or framework code would do that for you.

tyranron: Yeah, I've realized that `.clone()` happens because of `LocalBoxFuture` having lifetime parameter. And as nor Rust supports `'self` lifetimes, neither we have lifetime parameter in type signature, we have to use only `'static` there, so that's a definitely requires clone.

PR #65, opened 2019-11-14
Fix code style

JohnTitor:
Fix clippy warnings and run `cargo fmt`
Particularly, we should fix `mutable_borrow_reservation_conflict`.


actix-server/src/ssl/openssl.rs
```diff
@@ -92,7 +92,7 @@ impl<T: AsyncRead + AsyncWrite + Unpin + 'static, P> Service for OpensslAcceptor
                 let acc = acc;
                 tokio_openssl::accept(&acc, io).await
             }
-                .boxed_local::<'static>(),
+            .boxed_local::<'static>(),
```
fafhrd91: I think `static could be removed

JohnTitor: Exactly, removed the lifetime

PR #69, opened 2019-12-02
Fix for non Unix

Dispersia:
Warnings were turned into errors, but windows doesn't construct 3 of the variants.


fafhrd91:
Thanks!

Issue #70, opened 2019-12-03
Switch AppVeyor to GitHub Actions


JohnTitor:
It'd be nice to run CI on Windows to avoid build failures. And I think we can try it with GitHub Actions to make CI fast.

JohnTitor:
Current status:
- macOS: failed in test
- Windows: failed to install `openssl-sys`

klmallory:
adding to this, I am unable to build in alpha.2 on windows...

error[E0432]: unresolved import `self::unix`
  --> C:\Users\Me\.cargo\registry\src\github.com-1ecc6299db9ec823\actix-rt-1.0.0-alpha.2\src/lib.rs:56:19
   |
56 |     pub use self::unix::*;
   |                   ^^^^ could not find `unix` in `self`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0432`.
error: could not compile `actix-rt`.
warning: build failed, waiting for other jobs to finish...
error: build failed

JohnTitor:
@klmallory It has been fixed in `1.0.0-alpha.3`.

fafhrd91:
ci action works

PR #71, opened 2019-12-03
Use GitHub Actions

JohnTitor:
Run CI on Windows and macOS.
It's experimental so I don't remove AppVeyor for now.
CC: #70 
See the check suite: https://github.com/JohnTitor/actix-net/actions

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,58 @@
+name: CI
+
+on: [push, pull_request]
+
+jobs:
+  build_and_test:
+    strategy:
+      fail-fast: false
+      matrix:
+        toolchain:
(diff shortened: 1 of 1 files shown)


fafhrd91:
Does it support codecov integration?

JohnTitor:
I'm not sure, but I guess we can do it. In that case, we can also migrate from Travis.

fafhrd91:
as I understand we still going to use linux travis ci?

JohnTitor:
We can also run CI on Linux with GitHub Actions, so if we integrate codecov with Actions, I think Travis is unnecessary unless it has other parts to avoid using multiple CIs. (This PR doesn't contain Linux support and codecov integration though.)

PR #73, opened 2019-12-06
Stop running tests for all features on Actions

JohnTitor:
The current check suite on Windows is unhelpful. Until we come up with the way to compile `rust-openssl` on Windows, I'd like to stop running tests for all features on Actions.


fafhrd91:
Let’s wait a bit. I’d like to refactor tests

fafhrd91:
will do refactoring later

PR #77, opened 2019-12-18
rename router to actix-router

yjhmelody:
And remove warning


fafhrd91:
why?

yjhmelody:
Because crate name is  actix-router ：）

fafhrd91:
so what? router name is used from the beginning.

PR #78, opened 2019-12-18
Run tests for all features as possible on Actions

JohnTitor:
Run tests for all features on windows-msvc and macOS.
Somehow I couldn't compile openssl-sys on windows-gnu, so it's an exception.

The check suite is here: https://github.com/JohnTitor/actix-net/commit/049e9bca3821b6d3546b2bab27c5b245a6768506/checks?check_suite_id=364356395

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -2,6 +2,9 @@ name: CI
 
 on: [push, pull_request]
 
+env:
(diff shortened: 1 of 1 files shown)


JohnTitor:
> should we enable actions for actix-web as well?

I think it can be an alternative for AppVeyor there enough. I will test on my fork.

PR #81, opened 2019-12-27
actix-connect: Update trust-dns-resolver to a release

ignatenkobrain:
It is much better to use release version instead of pinned pre-release.

Signed-off-by: Igor Gnatenko <i.gnatenko.brain@gmail.com>

### DIFF
--- a/actix-connect/Cargo.toml
+++ b/actix-connect/Cargo.toml
@@ -41,8 +41,7 @@ either = "1.5.2"
 futures = "0.3.1"
 http = { version = "0.2.0", optional = true }
 log = "0.4"
-trust-dns-proto = "=0.18.0-alpha.2"
-trust-dns-resolver = "=0.18.0-alpha.2"
+trust-dns-resolver = "0.18.0"
 
 # openssl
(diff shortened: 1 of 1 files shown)


ignatenkobrain:
Oh, I see why this is broken... @fafhrd91 do you plan to update to 0.18.0 anytime soon or should I try to do so myself?

JohnTitor:
I believe it's a good thing to try yourself :)

saghm:
Hello! I'm curious if there's any timeline for having this issue fixed. We just had to bump our `trust-dns` dependencies in the [MongoDB driver](https://github.com/mongodb/mongo-rust-driver/pull/124) due to the requirement of `=0.18.0-alpha2` conflicting with our `0.18.0`. While I understand that there are breaking changes here that would take some time to fix, I think it's worth pointing out that other crates might also be requiring the stable version of `0.18.0` and aren't able to upgrade to `0.19.0` as easily as we were able to; pinning to an alpha version shifts a lot of the burden of compatibility to other maintainers.

PR #86, opened 2020-01-12
Add support for tokio tracing for actix Service.

avranju:


JohnTitor:
I'm not sure this should be in actix-net, not third-party crate. CC @fafhrd91

fafhrd91:
I asked Raj to add it to actix-net

actix-tracing/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+
+use actix_service::{apply, IntoServiceFactory, Service, ServiceFactory, Transform};
+use futures_util::future::{ok, Either, Ready};
+use tracing_futures::{Instrument, Instrumented};
+
+/// Trait representing a type that is able to produce spans for a given value.
+///
+/// The generic argument `T` is typically the request type of a `Service`
+/// implementation.
+pub trait MakeSpan<T> {
```
fafhrd91: what is the reason for trait here? why not just simple `Fn(&T) -> Option<Span>`?

actix-tracing/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// Creates a new span for the given request value.
+    ///
+    /// By default, this function returns `None`.
+    fn span_for(&self, _req: &T) -> Option<tracing::Span> {
```
fafhrd91: should be required method, no reason for default impl

actix-tracing/src/lib.rs
```diff
@@ -0,0 +1,302 @@
...
+///     web_service,
+///     |req: &Request| Some(span!(Level::INFO, "request", req.id))
+/// );
+/// ```
+///
+/// [`MakeSpan`]: trait.MakeSpan.html
+pub fn trace<S, U, T>(
+    service_factory: U,
+    make_span: T,
+) -> impl ServiceFactory<
```
fafhrd91: it won't be possible to use `+Clone`  with `impl Trait` here

actix-tracing/src/lib.rs
avranju: Made it a trait in the anticipation that we might have more span making functions in the future (for `poll_ready` for instance) and provided a blanket impl for `Fn(&T) -> Option<Span>`. But I don't feel strongly about this. We can always have a trait later if we think its appropriate. Will change.

actix-tracing/src/lib.rs
avranju: Agree.

actix-tracing/src/lib.rs
avranju: Not sure I understand this. I was able to add a `Clone` bound on the `Fn`.

fafhrd91: Resulted factory wouldn’t have clone even if input service factory would have clone

fafhrd91: Let’s return `ApplyTransform` type instead of impl ServiceFactory

avranju: Changed the return type to `ApplyTransform`.

Still don't understand the `Clone` comment. The `Clone` bound is on the closure and not on the factory returned from `trace`.

fafhrd91: but if you want to make clonable service factory, it would be not possible with impl Trait

fafhrd91:
thanks!

cdbattags:
Any documentation yet for using this with `actix-web`?

PR #89, opened 2020-01-21
Optimize vector fill in builder.

zero-systems:
Original code uses `Vec::new` and then adds entries multiple times through `push`. That can cause unnecessary memory reallocation and capacity checks. There commits should fix it using range with map.


JohnTitor:
Thanks!

PR #90, opened 2020-01-29
Tweak GitHub Actions

JohnTitor:
- Separate GHA config files
- Fix installation of OpenSSL on Windows builders
- Run full tests on windows-gnu
- Remove AppVeyor config
- Fix/suppress Clippy warnings


.github/workflows/clippy.yml
```diff
@@ -0,0 +1,17 @@
...
+name: Clippy Check
+jobs:
+  clippy_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: nightly
+          components: clippy
```
tyranron: Do we really need `nightly` for Clippy? I haven't seen that some unstable Clippy features were used. So I suggest to go with `stable` one.

actix-testing/src/lib.rs
```diff
@@ -1,6 +1,6 @@
 //! Various helpers for Actix applications to use during testing.
 #![deny(rust_2018_idioms, warnings)]
-#![allow(clippy::type_complexity)]
+#![allow(clippy::type_complexity, clippy::needless_doctest_main)]
```
tyranron: We also may harden `rustc` linter requirements, like these ones:
```rust
#![deny(
    missing_debug_implementations,
    nonstandard_style,
    rust_2018_idioms,
    trivial_casts,
    trivial_numeric_casts,
    unsafe_code
)]
#![warn(
    deprecated_in_future,
    missing_copy_implementations,
    missing_docs,
    unreachable_pub,
    unused_import_braces,
    unused_labels,
    unused_lifetimes,
    unused_qualifications,
    unused_results
)]
```

But it's better to do in a separate PR.

.github/workflows/clippy.yml
JohnTitor: I chose nightly's Clippy since it has new lints and the fix for false positives/negatives.

actix-testing/src/lib.rs
JohnTitor: It sounds good!

.github/workflows/clippy.yml
tyranron: But consider that it may accidentally introduce new unstable lints, which doesn't work well. At my job we had quite a pain with that, as CI sporadically failed. If it's not a big problem - I'm OK with that, anyway.

JohnTitor: `clippy check` only fails when the code has errors (i.e. denied lints detect something) and we can ignore it unless the check is set as required status so it shouldn't be a pain here, I think.

PR #94, opened 2020-02-06
Allow awaiting for futures spawned on current arbiter

Jonathas-Conceicao:
This PR includes essentially two changes: (1) a basic integration test is added, the test validate runtime execution semantics using `tokio::time::delay_for` and elapsed time to ensure futures are being driven to completion; (2) a `Arbiter::local_join` function was added, I'll focus the rest of this description about it.

# Motivation
When we're is doing computation to a newly allocated `Arbiter`one can use `Arbiter::join`in order to wait for it's completion, however when futures are spawned to the current arbiter we can lose track of it's completion status. Some crates, like actix will [eventually spawn futures to the current Arbiter](https://github.com/actix/actix/blob/f5eda8c75b1404a6215a396f23cc36f820ea7ef3/src/context.rs#L107), and it seams that there's no way of asking the system rather any of them are still running.

# Solution
The added `Arbiter::local_join`, a associated function that will return a future that will only complete once all current spawned futures have completed.

# Implementation
Since we use `tokio::task::spawn_local` which returns a [JoinHandle](https://docs.rs/tokio/0.2.11/tokio/task/struct.JoinHandle.html) to the spawned future, it's pretty doable to keep track of future's progress (maybe even resolved value later on).

So now all spawned futures's JoinHandle are put into a buffer of spawned futures, when one asks for a `Arbiter::local_join` this buffer is cleared and a `Future` is returned that will be resolved once all spawned futures have have finished.

# Future work
I believe it would be viable to change the API so functions like `actix_rt::spawn` can have a return value and are responded with a handle that can be used to access it if the user so desires.

### DIFF
--- a/actix-rt/CHANGES.md
+++ b/actix-rt/CHANGES.md
@@ -4,6 +4,8 @@
 
 - Expose `System::is_set` to check if current system is running
 
+- Add `Arbiter::local_join` associated function to get be able to `await` for spawned futures
+
 ## [1.0.0] - 2019-12-11
 
 * Update dependencies
--- a/actix-rt/src/arbiter.rs
+++ b/actix-rt/src/arbiter.rs
@@ -15,10 +15,13 @@ use crate::system::System;
 
 use copyless::BoxHelper;
 
+pub use tokio::task::JoinHandle;
+
 thread_local!(
     static ADDR: RefCell<Option<Arbiter>> = RefCell::new(None);
     static RUNNING: Cell<bool> = Cell::new(false);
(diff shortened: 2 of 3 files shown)


actix-rt/CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Changes
 
+## [TBD] - [TBD]
+
+- Add `Arbiter::local_join` assiciated function to get be able to await for spawned futures
```
therealprof:
```suggestion
- Add `Arbiter::local_join` associated function to get be able to `await` for spawned futures
```

PR #99, opened 2020-02-22
Expose is_set

ailisp:
In a situation I need check if there's a currently running actix system, but not panic when there isn't. This pub crate function just works and seems not dangerous to expose it.

### DIFF
--- a/actix-rt/src/system.rs
+++ b/actix-rt/src/system.rs
@@ -79,8 +79,8 @@ impl System {
         })
     }
 
-    /// Set current running system.
-    pub(crate) fn is_set() -> bool {
+    /// Check current system is running.
+    pub fn is_set() -> bool {
         CURRENT.with(|cell| cell.borrow().is_some())
     }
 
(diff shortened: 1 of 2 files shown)


actix-rt/src/system.rs
```diff
@@ -79,8 +79,8 @@ impl System {
         })
     }
 
-    /// Set current running system.
-    pub(crate) fn is_set() -> bool {
+    /// Check current system is running.
```
Jonathas-Conceicao: /// Check if current system is running.

JohnTitor:
@Jonathas-Conceicao On some repositories (actix, actix-net, actix-web), merge should be done with care. And I didn't review this yet...

Jonathas-Conceicao:
@JohnTitor not sure what I did wrong here. But I will start to wait on your review before merging in the future.

JohnTitor:
@Jonathas-Conceicao Don't have to wait though, but you could ping me when API is changed, it'd be great! Otherwise, I may miss the change :)

PR #104, opened 2020-02-26
`actix-rt`: minimize `futures` dependencies to `futures-{channel,util}` with default features off

ErichDonGubler:
This knocks out a good 12 deps in a debug build. Yeehaw!

### DIFF
--- a/actix-rt/Cargo.toml
+++ b/actix-rt/Cargo.toml
@@ -18,6 +18,7 @@ path = "src/lib.rs"
 [dependencies]
 actix-macros = "0.1.0"
 actix-threadpool = "0.3"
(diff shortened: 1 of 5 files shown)


robjtede:
Just wondering, is it standard practice to still be importing the Future trait from `futures` crate?

therealprof:
@robjtede The futures re-exports `std::future::Future` and extends it. It might be possible to switch actix to user `std::future::Future` and async-std instead, but `tokio` will still pull in `futures`. They're two alternative ecosystems built around the same concepts and fundamental types.

JohnTitor:
Thanks @ErichDonGubler and @therealprof!

PR #113, opened 2020-03-05
WIP: Convert actix-service to use RefCell instead of UnsafeCell

dunnock:
Hi, this is first PoC implementation for gathering feedback to make sure I am on the right path.

For smooth conversion I have added AXCell, which is similar to former Cell but using RefCell under the hood. Also added couple of concurrency tests with delays on first and second service to validate that there are no 2 mutable borrows which would lead to panic with RefCell.

By benchmarks pipeline with new AndThen performing similar to UnsafeCell baseline implementation:
```
AndThen with UnsafeCell #2              time:   [54.157 ns 54.850 ns 55.605 ns]
AndThen with RefCell #2                 time:   [54.976 ns 55.624 ns 56.353 ns]                                    
Pipeline::and_then based Rc<RefCell>    time:   [54.696 ns 55.260 ns 55.866 ns]
```

Pls let me know if using this approach is ok?


dunnock:
FYI just submit issue about windows mingw build: https://github.com/numworks/setup-msys2/issues/29

JohnTitor:
Disabled mingw builders temporarily on master.

cdbattags:
I'm good with this approach.

Might be nice to encapsulate this module elsewhere if we do decide to go this way.

cetra3:
Looks good! I like the abstraction of `AXCell` but think the name might be a little confusing (I am definitely bikeshedding though!).

Shnatsel:
This not only removes unsafe code, this is also an important soundness fix. Is there anything that's holding up the merge? It'd be nice to get this landed and released sooner rather than later.

The approach looks good to me. `Rc<RefCell<T>>` is indeed how Actix originally functioned, before the introduction of the custom cell: https://github.com/actix/actix-net/commit/20b03a478063d6bb293a6636632f976113812e13

robjtede:
> Is there anything that's holding up the merge?

It’s still in draft and needs some perf measurements done so we can know the impact of any regression.

Also, my focus is on v3 release blockers which this isn’t. It could be a patch release later.

dunnock:
Sorry for delaying on this, I was dragged in some other works, but will try to push this further during next week.
Do you know why RefCell was changed to UnsafeCell in the first place?

Shnatsel:
>  Do you know why RefCell was changed to UnsafeCell in the first place?

No. The commit message merely reads "add custom cell".

I *suspect* the reason was performance - I believe it is possible to increment one counter instead of two on every access, if you place certain additional restrictions on the usage patterns. I believe this was the intent, although I cannot know for sure.

I do not expect this to have any meaningful effect on performance outside of microbenchmarks. The counters are not atomic and are placed on the same L1 cache line. The cost of performing two increments instead of one should be so trivial as not to be worth considering compared to even an L2 cache access. Or that's the theory, at least.

Shnatsel:
@dunnock this looks complete to me, any reason to keep this in draft?

This fixes a long-standing memory safety issue that allows obtaining several mutable references to the same data and get arbitrary memory corruption from there ([PoC that fails MIRI](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3b2cabc2ba83096418d55ecee5c71809)). Moreover, benchmarks demonstrating lack of performance impact are included. I'm surprised that 6 months later this is still not merged.

robjtede:
Adding this to [v3 blockers](https://github.com/actix/actix-web/issues/1554) list so I don't forget about it again.

dunnock:
I've run into issues with `AndThenApplyFn`, probably that was the reason why the `Cell` was introduced in the first place..
since  `b` is RefMut<T> now, not `T` as before, it does not let to destructure tuple in the following code:
```rust
                    let mut b = b.get_mut();
                    let fut = (&mut b.2)(res, &mut b.1);
```
Probably will need to split that tuple in the first place...

Shnatsel:
There is a [simpler patch](https://gist.github.com/Nemo157/b495b84b5b2f3134d19ad71ba5002066) to fix the soundness holes that replaces all occurrences of Actix cell with `Rc<RefCell<T>>`, the exact same thing that was used before the introduction of the custom cell. It passed tests when it was written (same age as this PR).

Any reason not to go with that right now to fix the memory issue? We can always think of a fancier solutions later.

robjtede:
Closing this since #158 merged. Appreciate everyones input on this important fix.

dunnock:
Cool, thanks for taking lead on this!

PR #121, opened 2020-03-16
Revert "Disable windows-mingw builder temporarily"

JohnTitor:
Reverts actix/actix-net#115


JohnTitor:
Seems the upstream issue is resolved, reverting.

PR #124, opened 2020-04-04
actix-rt: Add Arbiter::is_running helper and fix System::is_set doc 

Jonathas-Conceicao:
The doc for `System::is_set` states:

> Check if current system is running.

Which is not exactly right. This only checks if system is set, i.e., if it has been started at some point. But it will still be true when the system eventually stops. The following test would panic if used with current API.

```Rust
#[test]
fn start_and_stop() {
    actix_rt::System::new("start_and_stop").block_on(async move {
        assert!(
            actix_rt::System::is_set(),
            "System doesn't seem to have started"
        );
    });
    assert!(
        !actix_rt::System::is_set(),
        "System doesn't seem to have stopped"
    );
}
```

Therefore I have updated it's description to represent what it actually menas. I've added `Arbiter::is_running` function to check if event loop is running for the current thread which does work with the test showed here.

### DIFF
--- a/actix-rt/tests/integration_tests.rs
+++ b/actix-rt/tests/integration_tests.rs
@@ -1,6 +1,20 @@
 use std::time::{Duration, Instant};
 
 #[test]
+fn start_and_stop() {
+    actix_rt::System::new("start_and_stop").block_on(async move {
+        assert!(
+            actix_rt::Arbiter::is_running(),
+            "System doesn't seem to have started"
+        );
+    });
+    assert!(
+        !actix_rt::Arbiter::is_running(),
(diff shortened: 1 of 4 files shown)


actix-rt/tests/integration_tests.rs
```diff
@@ -1,5 +1,19 @@
 use std::time::{Duration, Instant};
 
+#[test]
+fn start_and_stop() {
+    actix_rt::System::new("start_and_stop").block_on(async move {
+        assert!(
+            actix_rt::Arbiter::is_running(),
+            "System doesn't seam to have started"
```
robjtede, line 8: typo seam => seem

PR #127, opened 2020-04-08
Forward actix_rt::test arguments to test function.

rubdos:
Previously,

```rust
#[actix_rt::test]
async fn foo(_a: u32) {}
```

would compile to

```rust
#[test]
fn foo() {/* something */}
```

This patches changes this behaviour to

```rust
#[test]
fn foo(_a: u32) {/* something */}
```

by simply forwarding the input arguments.

This allows any test fixture library (e.g. `rstest`, cfr. https://github.com/la10736/rstest/issues/85) to integrate with actix::test.

Please let me know what you think of this approach, and whether you have another suggestion!

### DIFF
--- a/actix-macros/src/lib.rs
+++ b/actix-macros/src/lib.rs
@@ -58,6 +58,7 @@ pub fn test(_: TokenStream, item: TokenStream) -> TokenStream {
     let input = syn::parse_macro_input!(item as syn::ItemFn);
 
     let ret = &input.sig.output;
+    let inputs = &input.sig.inputs;
     let name = &input.sig.ident;
     let body = &input.block;
     let attrs = &input.attrs;
@@ -81,7 +82,7 @@ pub fn test(_: TokenStream, item: TokenStream) -> TokenStream {
     let result = if has_test_attr {
         quote! {
             #(#attrs)*
-            fn #name() #ret {
+            fn #name(#inputs) #ret {
                 actix_rt::System::new("test")
                     .block_on(async { #body })
             }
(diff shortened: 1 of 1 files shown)


rubdos:
I notice coverage went down by 0.02%. If you want me to cover some more code, I can add a patch like this:

```
diff --git a/actix-macros/Cargo.toml b/actix-macros/Cargo.toml
index eb72e52..e59dac1 100644
--- a/actix-macros/Cargo.toml
+++ b/actix-macros/Cargo.toml
@@ -19,3 +19,4 @@ syn = { version = "^1", features = ["full"] }

 [dev-dependencies]
 actix-rt = { version = "1.0.0" }
+rstest = "0.6"
diff --git a/actix-macros/src/lib.rs b/actix-macros/src/lib.rs
index ffd8147..86d448a 100644
--- a/actix-macros/src/lib.rs
+++ b/actix-macros/src/lib.rs
@@ -53,6 +53,18 @@ pub fn main(_: TokenStream, item: TokenStream) -> TokenStream {
 ///     assert!(true);
 /// }
 /// ```
+///
+/// It also forwards arguments to tests, such that it works with fixture libraries.
+///
+/// ```no_run
+/// use rstest::*;
+///
+/// #[actix_rt::test]
+/// #[rstest]
+/// async fn my_test(fix: Fix) {
+///     assert!(true);
+/// }
+/// ```
 #[proc_macro_attribute]
 pub fn test(_: TokenStream, item: TokenStream) -> TokenStream {
     let input = syn::parse_macro_input!(item as syn::ItemFn);
```

It does however require a fixture libary to "really" cover the new function.

JohnTitor:
> I notice coverage went down by 0.02%. If you want me to cover some more code, I can add a patch like this:

I'd like to ignore the coverage here.

PR #133, opened 2020-04-24
add macro compile tests

robjtede:
Add compile tests using `trybuild`.


robjtede:
Not been having fun trying to wrangle compiletest.

robjtede:
thanks to some swift changes to trybuild this PR should be ready now that windows CI failures have been addressed.

PR #137, opened 2020-05-14
chore: Update tokio-util to 0.3

ignatenkobrain:


ignatenkobrain:
Oh, I just realized that this is breaking change, and requires some actual change to the code (not sure why I did not notice this before sending PR).

This would be appreciated anyway, if you could do it.

cdbattags:
We can target 3.0 for this change.

ignatenkobrain:
@cdbattags would be much appreciated, thanks!

PR #150, opened 2020-06-15
Expose fn build_async in builder

hengchu:
Currently, the `System::run_in_tokio` method uses `build_async`, but `build_async` itself is not exposed. This hides certain customization options on the builder if a user would like to provide a tokio local set manually. Would it be OK to completely expose `build_async`?

### DIFF
--- a/actix-rt/src/builder.rs
+++ b/actix-rt/src/builder.rs
@@ -56,7 +56,7 @@ impl Builder {
     /// Create new System that can run asynchronously.
     ///
     /// This method panics if it cannot start the system arbiter
-    pub(crate) fn build_async(self, local: &LocalSet) -> AsyncSystemRunner {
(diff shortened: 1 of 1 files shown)


robjtede:
> certain customization options on the builder

Such as? We can probably come up with alternate methods on the builder that don't expose this internal struct.

hengchu:
For now there is just this one configuration `stop_on_panic`, but there might be more in the future?

https://github.com/actix/actix-net/blob/7140c04c44842d15dda9caab3b8461f0c15adaa3/actix-rt/src/builder.rs#L44-L47

hengchu:
The general idea is that `System::Builder` may get more customization features in the future, and whatever interface we come up with will probably have to duplicate those features if this internal struct is not exposed.

But I understand this is probably more suitable as an issue for design discussion. Closing the PR for now.

PR #158, opened 2020-07-19
Remove unsound custom Cell

Shnatsel:
Right now the custom cell is unsound because `Cell::get_mut()` uses `Rc::as_ref()` to obtain a reference to the underlying data, which does not distinguish between mutable and immutable references, and so does not guarantee uniqueness. Thus it is possible to obtain several mutable references to the same memory location by calling `Cell::get_mut()` repeatedly:

```rust
let mycell = Cell::new(vec![1,2,3]);
let ref1 = mysell.get_mut();
let ref2 = mysell.get_mut(); // obtained a second mutable reference; UB starts here
```

This unsoundness can even be triggered from the public API (see [PoC that fails MIRI](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3b2cabc2ba83096418d55ecee5c71809)). It may result in pretty much arbitrary memory corruption, most likely a use-after-free.

This PR removes the unsound custom cell and replaces it with `Rc<RefCell<T>`. This reverts the code to the way it was before the introduction of the custom cell in 20b03a478063d6bb293a6636632f976113812e13 
I expect performance to take a slight hit, but the difference not to be noticeable outside of microbenchmarks. `cargo bench` shows no difference between this branch and master.

The code is heavily based on the patch by @nemo157 https://gist.github.com/Nemo157/b495b84b5b2f3134d19ad71ba5002066, all I did was resolve conflicts with latest master.

This PR supersedes #113 which has been stuck for 6 months and didn't fix the problem entirely, since the unsound custom cell was still in use.


robjtede:
Thanks so much for expiditing this. Can you mention fix in the changelog.

Shnatsel:
> Can you mention fix in the changelog.

Done.

Shnatsel:
Error message from failed check is: 

>   Unable to update registry `https://github.com/rust-lang/crates.io-index`

So looks like it's just flaky and needs to be retried.

Shnatsel:
CI is green now.

Shnatsel:
Thanks for the quick review and merge!

Please let me know once a version incorporating these changes is released. I'd like to open a RustSec advisory so that anyone interested in security would know to update.

Shnatsel:
Oh wait, there's _another_ copy if this Cell implementation in the repo. I'll open an issue about that one.

PR #161, opened 2020-07-20
actix-utils: Remove unsound custom Cell as well

### LINKED ISSUE #160: Custom Cell implementation is unsound
Right now there is no mechanism in `Cell` to track whether a mutable reference to the data is already acquired. Thus it is possible to obtain several mutable references to the same memory location by calling `Cell::get_mut()` repeatedly:

```rust
let mycell = Cell::new(vec![1,2,3]);
let ref1 = mysell.get_mut();
let ref2 = mysell.get_mut(); // obtained a second mutable reference; UB starts here
```

This may result in pretty much arbitrary memory corruption, most likely a use-after-free. Even though no code internal to Actix makes two obvious calls to `get_mut()` in a row, this behavior has been shown to be exploitable from the public API (see [PoC that fails MIRI](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3b2cabc2ba83096418d55ecee5c71809)).

PR #158 has removed one instance of this Cell and all uses of it. However, there is [another copy of it](https://github.com/actix/actix-net/blob/master/actix-utils/src/cell.rs) in the repository in `actix-utils` crate, and there are 23 calls to `.get_mut()`.

The obvious fix is to replace all uses of custom `Cell<T>` with `Rc<RefCell<T>>` (the way it was before the introduction of a custom cell in 20b03a478063d6bb293a6636632f976113812e13), like it was done in #158.

JohnTitor:
## PR Type

Bug Fix/Refactor

## PR Checklist

## Overview

The changes are the almost same as #158, but it's for actix-utils.
Fixes #160
cc @Shnatsel


Shnatsel:
Looks good to me on the surface. I can't claim to understand that's going on, though.

Thanks for the swift fix!

PR #165, opened 2020-07-23
rework actix-threadpool

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

`actix-threadpool` would spawn a fixed number of threads in a static pool and stay for the life of whole actix-app lifetime. I feel this is an overkill for majority use case and may bring up confusion.

This PR reworks the pool to use a dynamic strategy with the following behavior:
- Rename the worker thread name from "actix-web" to "actix-threadpool-worker".
- Start with 1 idle thread and when more jobs goes in it spawn more threads until hitting the upper limit.
- Any pool thread kept idle for 5 minutes would de spawn it self.

This PR is still not complete and need some help on tests and benchmarks.

### DIFF
--- a/actix-threadpool/CHANGES.md
+++ b/actix-threadpool/CHANGES.md
@@ -1,5 +1,7 @@
 # Changes
 
+* Rework pool's gut.
+
 ## [0.3.3] - 2020-07-14
 
 ### Changed
--- a/actix-threadpool/Cargo.toml
+++ b/actix-threadpool/Cargo.toml
@@ -24,6 +24,7 @@ jian_rs = { git = "https://github.com/fakeshadow/jian_rs"}
 lazy_static = "1.3"
 log = "0.4"
 num_cpus = "1.10"
+parking_lot = "0.11.0"
 
 [dev-dependencies]
 actix-rt = "1"
--- a/actix-threadpool/Cargo.toml
+++ b/actix-threadpool/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "actix-threadpool"
-version = "0.3.3"
+version = "0.4.0"
 authors = ["Nikolay Kim <fafhrd91@gmail.com>"]
 description = "Actix thread pool for sync code"
 keywords = ["actix", "network", "framework", "async", "futures"]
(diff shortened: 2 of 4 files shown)


cdbattags:
@fafhrd91 I'd love your context on design decisions here. This kind of flys in the face of performance and what you had attempted at the beginning for speed.

@fakeshadow any chance we could change default behavior but still have old behavior as an option?

fakeshadow:
From my limit tests I barely noticed any performance regress.

There are more configures to work with. e.g: 
```rust
let num = (num_cpus::get() * 5).to_string();
std::env::set_var("ACTIX_THREADPOOL", num.as_str());
std::env::set_var("ACTIX_THREADPOOL_MIN",  num.as_str());
std::env::set_var("ACTIX_THREADPOOL_TIMEOUT", "0");
```
This would have an identical behavior as the current pool.

fakeshadow:
I added a benchmark [here](https://github.com/fakeshadow/jian_rs/tree/master/benches) to mimic the default behavior of actix-threadpool with the new one.
On my machine the result are at the same level.

robjtede:
I think that apprehensions about replacing such an important part of the system could be quelled by releasing this as a separate crate that, once battle tested, could replace the original implementation. What do you think about making this a new `actix-dynamic-threadpool` package or something?

fakeshadow:
I like the idea of a separate package or just leave this poll request as draft for now.
The dynamic pool can be easily import from my fork of actix-net in `Cargo.toml`. And I hope people can use it and give some feed back.

robjtede:
I think my ideal plan would be to release this as a separate crate, with a dependency on a published version of your underlying jian_rs, and maybe add a feature flag to actix-web (post-v3) to replace the web::block call at developers choice. This would give more visibility and use to it that having to use a -net fork.

fakeshadow:
Sounds a good plan. I can rename the project to `acitx-dynamic-threadpool` and remove `jian_rs` all together. If that's acceptable.

robjtede:
> remove jian_rs all together

If I understand, that's the lib driving this new implementation, in which case I personally wouldn't have an issue using it as a dependency (as long as it's published). It's your work and you deserve to be credited for it. What did you have in mind by "removing" it?

fakeshadow:
`jian_rs` have no essential dependency besides std. So it would be perfectly fine to copy paste the code to actix-threadpool and merge it into one separate package.

robjtede:
Sure, inlining the package would be okay too.

fakeshadow:
Ok. would do that and I believe this PR can be closed.

PR #167, opened 2020-07-23
assert workers greater than zero

### LINKED ISSUE #166: manually setting workers(0) will actually spin 0 thread
Just a little thing, I noticed. at server building when you set .workers(0) then it will actually spin 0 threads and no request goes though. My expectation was that by setting 0, it would load whatever the default value for workers would be. Is it a bug or desired behavior? If bug, I'd be glad to send a PR.
right now with: actix-web "3.0.0-beta.1"

adrian-wechner:
## PR Type
Feature

## PR Checklist

## Overview
Before setting workers(0) will actually spin 0 threads. With this change, when setting 0 it defaults back to available logical cpus using num_cpus::get(). Use case, is having a configuration system which might request the default value and can do this by passing 0 as the worker parameter.

Closes #166

### DIFF
--- a/actix-server/CHANGES.md
+++ b/actix-server/CHANGES.md
@@ -1,5 +1,11 @@
 # Changes
 
+## Unreleased
+
+### Changed
+
+* workers(0) defaults to num_cpus::get()
+
 ## [1.0.3] - 2020-05-19
 
 ### Changed
--- a/actix-server/CHANGES.md
+++ b/actix-server/CHANGES.md
@@ -4,7 +4,7 @@
 
 ### Changed
 
-* workers(0) defaults to num_cpus::get()
+* workers must be greater than 0
 
 ## [1.0.3] - 2020-05-19
 
--- a/actix-server/src/builder.rs
+++ b/actix-server/src/builder.rs
@@ -74,10 +74,8 @@ impl ServerBuilder {
     /// By default server uses number of available logical cpu as workers
     /// count. Explicitly setting 0 will also default.
     pub fn workers(mut self, num: usize) -> Self {
-        self.threads = match num {
-            0 => num_cpus::get(),
-            _ => num,
-        };
+        assert_ne!(num 0, "workers must be greater than 0");
+        self.threads = num;
         self
     }
 
--- a/actix-server/src/builder.rs
+++ b/actix-server/src/builder.rs
@@ -72,9 +72,12 @@ impl ServerBuilder {
     /// Set number of workers to start.
     ///
     /// By default server uses number of available logical cpu as workers
-    /// count.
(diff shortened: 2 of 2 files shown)


cetra3:
I feel like a better change here would be to panic if it's not above 0 rather than allowing that to be set, rather than this behaviour.

It's a little too "magical" for my liking!

fakeshadow:
I too feel an assert would be a better way to handle this.

robjtede:
From the original issue:

> I use it personally in our configuration system, that by setting 0 for the workers config, that we translate it into the system's cpu count.

Having dabbled in devops, I do think there's value in being *able* to use an env var to 0 and allow that to mean "take the CPU count". What do folks think about this as a proposal:

```rust
pub enum WorkerCount {
    Fixed(usize),
    NumCpus,
}

impl WorkerCount {
    fn count(&self) -> usize {
        match self {
            WorkerCount::Fixed(num) => *num,
            WorkerCount::NumCpus => num_cpus::get(),
        }
    }
}

impl From<usize> for WorkerCount {
    fn from(num: usize) -> WorkerCount {
        WorkerCount::Fixed(num)
    }
}

impl HttpServer {
    pub fn workers(self, num: impl Into<WorkerCount>) -> Self {
        let workers: WorkerCount = num.into();
        let workers = workers.count();
        assert!(workers > 0);

        // ....
    }
}
```

This is akin to the the KeepAlive options for HttpServer which can accept KeepAlive::OS in addition to a numeric fixed keep alive value, and solves the problem of being able to declare in the environment to use num_cpus more easily (no conditional function calls, only conditional variable assignments).

See playground for example:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=bcebe9c88927917a30891053c5f5fff1

cetra3:
Yep, that will allow you to include either a number or to specify the default.

It would be great though if there wasn't a way to construct the `WorkerCount` with `0` as the number.  I.e, enforce at compile time.  I'm not sure how that could be done, maybe an assert on the `From` impl? But then they could always just manually create the `WorkerCount::Fixed`

robjtede:
`NonZeroUsize` exists but would still require runtime checks. I think the assert in server initialization is enough.

robjtede:
Want to get a version cut pretty soon. What do you think of this discussion @adrian-wechner ?

adrian-wechner:
Hi there, personally I still think the best thing to do, is to reset to default as it feels to me the most natural thing to do in that moment. But I'd be equally fine with any other solution as any other soluton would require to check by myself to default. Maybe a simple warning to the log is the better solution? This way everybody can still do whatever, nothing breaks, and if debug is enabled at least is getting some retro. What do you think of that?

PR #170, opened 2020-08-03
Document accept loop

adwhit:
## PR Type

Refactor

## PR Checklist

## Overview

The core logic of `actix-server` is convoluted and mostly undocumented. Whilst attempting to track down bugs in my own code I dived in and tried to add comments to explain some of what is going on within the core accept loop.

By no means is this complete, and I cannot be certain my comments are even correct, but they might help future contributors. There are very few actual code changes, however I have renamed a few variables and function, but nothing public, and only in trivial ways. If any reviewers can point out extra details or things I got wrong, I will gladly update.

Incidentally, I think that large lots of this codebase could do with a proper refactor or rewrite. For example I was unable to get any workers into an `Unavailable` state, no matter how many concurrrent connections I tried, or whether I made the workers crash. I woder, since `tokio` will `catch_unwind` as of `0.2.0`, whether that is now dead code.

### DIFF
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -106,7 +106,7 @@ struct Accept {
     workers: Vec<WorkerClient>,
     srv: Server,
     timer: (mio::Registration, mio::SetReadiness),
-    next: usize,
+    next_worker_ix: usize,
     backpressure: bool,
 }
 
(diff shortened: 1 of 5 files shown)


robjtede:
Also could you add a brief note in the changelog.

fakeshadow:
Some of the commented code session are moved or renamed in another PR. I can pick this up later when it's settled.

fakeshadow:
I think this can be closed. 

A lot of documentation comments has beend added with changes and I feel actix-server code is more clean and self explained for most cases now.

There are more changes to come which would further render this PR outdated.

PR #173, opened 2020-08-06
actix-rt: Make the process of running System in existing Runtime more clear

popzxc:
## PR Type

Other (Usability enhancement)

## PR Checklist

## Overview

When I first tried to start the `actix-web` server using an existing `tokio` `Runtime`, I honestly found the interface a bit unfriendly.

In an attempt to make other folks' life easier I did the following:

- Added doc-comments with an example and a note about `Arbiter`s still using their own `Runtime` objects.
- Added a new method which is less flexible but simpler to use to attach a `System` to a given `Runtime`.

These changes aren't breaking and hopefully useful.

### DIFF
--- a/actix-rt/Cargo.toml
+++ b/actix-rt/Cargo.toml
@@ -25,5 +25,4 @@ smallvec = "1"
 tokio = { version = "0.2.6", default-features = false, features = ["rt-core", "rt-util", "io-driver", "tcp", "uds", "udp", "time", "signal", "stream"] }
 
 [dev-dependencies]
-futures = "0.3"
 tokio = { version = "0.2.6", features = ["full"] }
--- a/actix-rt/Cargo.toml
+++ b/actix-rt/Cargo.toml
@@ -23,3 +23,7 @@ futures-util = { version = "0.3.4", default-features = false, features = ["alloc
 copyless = "0.1.4"
 smallvec = "1"
 tokio = { version = "0.2.6", default-features = false, features = ["rt-core", "rt-util", "io-driver", "tcp", "uds", "udp", "time", "signal", "stream"] }
+
+[dev-dependencies]
+futures = "0.3"
+tokio = { version = "0.2.6", features = ["full"] }
--- a/actix-rt/src/system.rs
+++ b/actix-rt/src/system.rs
@@ -57,10 +57,58 @@ impl System {
         Self::builder().name(name).build()
     }
 
-    #[allow(clippy::new_ret_no_self)]
-    /// Create new system using provided tokio Handle.
+    /// Create new system using provided tokio `LocalSet`.
     ///
     /// This method panics if it can not spawn system arbiter
+    ///
+    /// Note: This method uses provided `LocalSet` to create a `System` future only.
+    /// All the [`Arbiter`]s will be started in separate threads using their own tokio `Runtime`s.
+    /// It means that using this method currently it is impossible to make `actix-rt` work in the `Runtime` other
+    /// than the provided by `tokio` 0.2 (e.g. provided by `tokio_compat`).
+    ///
+    /// [`Arbiter`]: struct.Arbiter.html
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use tokio::{runtime::Runtime, task::LocalSet};
+    /// use actix_rt::System;
+    /// use futures::future::try_join_all;
+    ///
+    /// async fn run_application() {
+    ///     let first_task = tokio::spawn(async {
+    ///         // ...
+    /// #        println!("One task");
+    /// #        Ok::<(),()>(())
+    ///     });
+    ///
+    ///     let second_task = tokio::spawn(async {
+    ///         // ...
+    /// #       println!("Another task");
+    /// #       Ok::<(),()>(())
+    ///     });
+    ///
+    ///     try_join_all(vec![first_task, second_task])
+    ///         .await
+    ///         .expect("Some of the futures finished unexpectedly");
+    /// }
+    ///
+    ///
+    /// let mut runtime = tokio::runtime::Builder::new()
+    ///     .core_threads(2)
+    ///     .enable_all()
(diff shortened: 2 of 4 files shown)


actix-rt/src/system.rs
```diff
@@ -57,10 +57,58 @@ impl System {
...
-    #[allow(clippy::new_ret_no_self)]
-    /// Create new system using provided tokio Handle.
+    /// Create new system using provided tokio `LocalSet`.
     ///
     /// This method panics if it can not spawn system arbiter
+    ///
+    /// Note: This method uses provided `LocalSet` to create a `System` future only.
+    /// All the [`Arbiter`]s will be started in separate threads using their own tokio `Runtime`s.
+    /// It means that using this method currently it is impossible to make `actix-rt` work in the `Runtime` other
+    /// than the provided by `tokio` 0.2 (e.g. provided by `tokio_compat`).
```
robjtede, line 67: Reword "other than the provided".

IMO, it's also note clear what in `tokio_compat` you're referring to.

actix-rt/Cargo.toml
```diff
@@ -23,3 +23,7 @@ futures-util = { version = "0.3.4", default-features = false, features = ["alloc
 copyless = "0.1.4"
 smallvec = "1"
 tokio = { version = "0.2.6", default-features = false, features = ["rt-core", "rt-util", "io-driver", "tcp", "uds", "udp", "time", "signal", "stream"] }
+
+[dev-dependencies]
+futures = "0.3"
```
robjtede, line 28: looks like you're only using try_join_all from this which is included in the already used `futures-util`

actix-rt/src/system.rs
```diff
@@ -71,6 +119,76 @@ impl System {
...
+    ///     .core_threads(2)
+    ///     .enable_all()
+    ///     .threaded_scheduler()
+    ///     .build()
+    ///     .unwrap();
+    ///
+    /// let rest_operations = run_application();
+    /// System::attach_to_tokio(runtime, "actix-main-system", rest_operations);
+    /// ```
+    pub fn attach_to_tokio<Fut, R>(
```
robjtede, line 177: I like the idea of this function, making it much simpler to host actix in an existing tokio runtime.

Are there alternate designs for the `rest_operations` idea worth discussing? I seem to recall in the past just `join`ing and awaiting the actix LocalSet with any other long running tasks. This approach feels more opinionated and less extensible.

actix-rt/src/system.rs
```diff
@@ -71,6 +119,76 @@ impl System {
...
+        mut runtime: tokio::runtime::Runtime,
+        name: impl Into<String>,
```
robjtede, line 179: Can we put name first, as with `run_in_tokio`

actix-rt/src/system.rs
popzxc: Well, I've attempted several other approaches, but found them not as convenient because of the following factors:
- `LocalSet` occupies current thread for `System` to run;
- Future, obtained via `LocalSet` is not `Send`. 

Thus, returning a future from this function (instead of blocking the runtime inside) makes the user-side code not really intuitive.

As an alternative approach, we may make `rest_operations` a function rather than a future, but IMHO it's less flexible: if you have an `async` function, you can get a future by simply calling it, and at the same time you can get a future without an additional function (e.g. using an `async { }` block).

Or do you have an idea how to get rid of the `rest_operations` argument without complications on the caller site?

actix-rt/src/system.rs
popzxc: Is it better now?

robjtede: yep, reads much better, thanks :)

popzxc:
@robjtede Can you please clarify how long it usually takes for a PR to get reviewed?

robjtede:
@popzxc I'm in favor of these changes. I'd like someone else from @actix/contributors team to chime in before we get it merged though.

popzxc:
I see, thanks! Let's wait then :)

actix-rt/src/system.rs
cdbattags: This feels like we're moving into a simpler Node.js style `<req, res>` or Koa's `ctx` and I'm all for it. I'd love to have Actix just be the web layer on top of default Tokio.

cdbattags: I think this is similar to what @fafhrd91 was describing too about the `actix-net` features.

popzxc:
@robjtede I don't want to be annoying, but it seems that nobody else want to review this PR.

Neopallium:
@popzxc I just looked over the changes.  I think it would be good to add a test for this new function.  Also the PR will need to be rebased against master.  Other then that it looks good.

Ping me later if you need another review.

popzxc:
> I think it would be good to add a test for this new function

@Neopallium  Isn't provided doc-test enough? If not, I'm not exactly sure what kind of test is required, could you provide an approximate form of how do you see it?

Neopallium:
I am not sure if doc-tests are run.  checking now.

popzxc:
Also after rebase there seems to be some inherited clippy warnings. Not sure if I have to fix them as part of this PR, as some of them don't even belong to the `actix-rt` crate.

Neopallium:
yeah.  I see those.  I don't think they should be fixed in this PR.  I can't merge this PR, but I can submit a review for you.

The Doc-tests seem to only check if the doc example code compiles.  A test would be good to make sure that attaching to an existing tokio-runtime still works (all normal tokio & actix tasks still run).  I am not sure if that would be easy to verify in a test.

Neopallium (approved):

Neopallium:
@popzxc I am fixing the clippy errors in PR #187 

@robjtede Is there anything else blocking this PR?

robjtede:
> Is there anything else blocking this PR?

@Neopallium nope, many thanks for your input here. Once the rebased checks pass we'll merge.

PR #175, opened 2020-08-08
prepare actix-service 1.0.6 release

robjtede:
May as well get the ball rolling on -net releases for inclusion in web v3. Though, given it's only a patch release, will be applicable to v2 web as well.

Made some corrections and improvements to the docs on Service too.

### DIFF
--- a/actix-service/Cargo.toml
+++ b/actix-service/Cargo.toml
@@ -1,12 +1,13 @@
 [package]
 name = "actix-service"
-version = "1.0.5"
+version = "1.0.6"
 authors = ["Nikolay Kim <fafhrd91@gmail.com>"]
-description = "Actix service"
-keywords = ["network", "framework", "async", "futures"]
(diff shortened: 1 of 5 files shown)


actix-service/Cargo.toml
```diff
@@ -1,12 +1,12 @@
...
+version = "1.0.6"
 authors = ["Nikolay Kim <fafhrd91@gmail.com>"]
-description = "Actix service"
-keywords = ["network", "framework", "async", "futures"]
+description = "Service trait and combinators for representing asynchronous request/response operations."
+keywords = ["network", "framework", "async", "futures", "service"]
 homepage = "https://actix.rs"
 repository = "https://github.com/actix/actix-net.git"
-documentation = "https://docs.rs/actix-service/"
+documentation = "https://docs.rs/actix-service"
```
JohnTitor, line 9: nit: Now, we can add the `readme` field :)

PR #138, opened 2020-05-18
Upgrade tokio utils to 0.3

robjtede:
## PR Type

Dependency Update

## PR Checklist
Check your PR fulfills the following:

- [ ] A changelog entry has been made for the appropriate packages.

## Overview

Updates tokio-utils to 0.3 in crates that need it and propagates type parameter changes. Also removes `actix-ioframe` as discussed.

PR #179, opened 2020-08-16
solve framed integration with actix-http

robjtede:
## PR Type
Bug Fix / Refactor

## PR Checklist

## Overview
- Follow up #138 and #110.
- Primarily, changes made to `Framed` while attempting integration with `actix-http`.
- Renames some methods to better describe their use.
- Adds a method to get a pinned reference to the framed IO.
- Fix things that were setting off my spell checker.

## Breaking Changes
- Many. But a major bump was already necessary for codec and utils from #138.

Quick note: will address checklist as we go but this should probably be considered in tandem with actix/actix-web#1634

### DIFF
--- a/actix-utils/CHANGES.md
+++ b/actix-utils/CHANGES.md
@@ -4,8 +4,7 @@
 
 * Upgrade `tokio-util` to `0.3`.
 * Remove unsound custom Cell and use `std::cell::RefCell` instead, as well as `actix-service`.
-* Provide correctly spelled `LocalWaker::is_registered` method and deprecate the replaced.
-* Alias `framed` module as `dispatcher`.
+* Rename method to correctly spelled `LocalWaker::is_registered`.
 
 ## [1.0.6] - 2020-01-08
 
--- a/actix-utils/CHANGES.md
+++ b/actix-utils/CHANGES.md
@@ -4,6 +4,8 @@
 
 * Upgrade `tokio-util` to `0.3`.
 * Remove unsound custom Cell and use `std::cell::RefCell` instead, as well as `actix-service`.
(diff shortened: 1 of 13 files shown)


robjtede:
Not entirely sure why the 1.39 build is failing. There does appear to be a 1.40 feature in the crate mentioned, they do claim MSRV of 1.39 but their CI is passing?

JohnTitor:
Because their MSRV check doesn't work correctly, submitted a fix: tokio-rs/tracing/pull/934
So, we also should bump MSRV to 1.40.0 (anyway it's fine for me since actix frends are going to 1.42.0 :) ).

JohnTitor:
Btw, almost changes look great to me, I'll take a more look later about the `Framed` change.

actix-utils/CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 * Upgrade `tokio-util` to `0.3`.
 * Remove unsound custom Cell and use `std::cell::RefCell` instead, as well as `actix-service`.
+* Provide correctly spelled `LocalWaker::is_registered` method and deprecate the replaced.
```
JohnTitor, line 7: "deprecate" here is not correct for me, I presume it's an oversight in e5b1aebc50d4fd185a71895b450beefbecb7d907.

JohnTitor (approved):
Overall looks good! Left one nit.

actix-utils/CHANGES.md
robjtede: ahh good spot, yeah i had the change in to keep the old version but realised that we had breaking changes anyway

PR #182, opened 2020-08-18
prepare codec and utils betas

robjtede:


JohnTitor (approved):
👍, the failure is due to https://github.com/alexcrichton/curl-rust/issues/352, let's ignore here.

PR #188, opened 2020-09-11
prepare server release 1.0.4

robjtede:
## PR Type
Release

## PR Checklist
- [x] Tests for the changes have been run against actix-web.

## Overview
actix-server v1.0.3 uses actix-codec 0.2.0 so this patch release can remove that (unsafe) dependency.

While technically "breaking", I feel that the change ensuring that worker counts are greater than 0 is an acceptable one since a server is no good without at least 1 worker. As a quick check, I did a search of GitHub for `.workers(0)` in Rust files and came back with nothing. Thoughts from @actix/contributors are valuable here.

### DIFF
--- a/actix-server/CHANGES.md
+++ b/actix-server/CHANGES.md
@@ -1,216 +1,128 @@
 # Changes
 
-## Unreleased
+## Unreleased - 2020-xx-xx
 
-### Changed
 
-* workers must be greater than 0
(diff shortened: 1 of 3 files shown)


JohnTitor (approved):

Neopallium (approved):

Neopallium:
Looks good.

PR #190, opened 2020-09-13
move and update server+tls examples

robjtede:
## PR Type
Docs

## PR Checklist

## Overview
A start on making these lower level primitives more accessible. Up until now the examples folder has not been part of any package and therefore hasn't been compile checked in CI.

This PR updates and documents the examples as well as generates a new key/cert pair for the TLS example. (Plus a couple docs tweaks that aren't important enough for changelog entries.)


JohnTitor (approved):
> Up until now the examples folder has not been part of any package and therefore hasn't been compile checked in CI.

Oh, I didn't notice that, looks good to me.

popzxc (approved):

PR #191, opened 2020-09-13
Add support of tokio-compat executor

### LINKED ISSUE #163: The Tokio runtime is not customizable for workers
## Expected Behavior
Using [`actix_rt::System::run_in_tokio`] I can run Actix with a custom Tokio runtime. Specifically, I want to use the [`tokio-compat`] runtime so I can support legacy Tokio 0.1 code.

## Current Behavior
Using [`actix_rt::System::run_in_tokio`] only runs the main "thread" or future under the custom runtime. Workers still use the default runtime ([link to relevant code](https://github.com/actix/actix-net/blob/f6c697a2dd706d8d981f439c9b77e60b982c16cd/actix-rt/src/arbiter.rs#L115)).

## Possible Solution
Provide a way to pass a runtime or runtime factory to Actix for its use.

## Steps to Reproduce (for bugs)

1. Clone and run https://github.com/Mcat12/actix-compat-runtime-crash
2. See that the spawned Tokio 0.1 and 0.2 futures ran (they print a message)
3. Go to http://127.0.0.1:8000/
4. See that the worker has panicked with "called `Result::unwrap()` on an `Err` value: SpawnError { is_shutdown: true }"

## Context
I need to use a crate which uses Tokio 0.1 (and futures 0.1). This crate sometimes spawns new futures (Hyper 0.12 client). This fails because there is no Tokio 0.1 executor.

Previous related issues:
- https://github.com/actix/actix-web/issues/913
- https://github.com/actix/actix-web/issues/1186

## Your Environment

* Rust Version (I.e, output of `rustc -V`): 1.44.1
* Actix Web Version: 2.0.0

[`actix_rt::System::run_in_tokio`]: https://docs.rs/actix-rt/1.1.1/actix_rt/struct.System.html#method.run_in_tokio
[`tokio-compat`]: https://docs.rs/tokio-compat/0.1.6/tokio_compat

popzxc:
## PR Type

Feature

## PR Checklist

## Overview

This PR adds experimental support of `tokio-compat` runtime.
In theory, it also provides a ground for making `actix-web` executor-agnostic, if that will ever be required.

For now, it makes `actix-rt` available for folks who have legacy code based on `tokio01`.

Resolves #163

### DIFF
--- a/actix-rt/src/system.rs
+++ b/actix-rt/src/system.rs
@@ -4,10 +4,10 @@ use std::io;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 use futures_channel::mpsc::UnboundedSender;
-use tokio::task::LocalSet;
 
 use crate::arbiter::{Arbiter, SystemCommand};
 use crate::builder::{Builder, SystemRunner};
+use crate::executor::LocalSet;
 
 static SYSTEM_COUNT: AtomicUsize = AtomicUsize::new(0);
 
@@ -143,6 +143,8 @@ impl System {
     /// # Examples
     ///
     /// ```
+    /// # #[cfg(feature = "tokio-executor")] // Test won't work for `tokio-compat-executor`.
+    /// # fn main() {
     /// use tokio::runtime::Runtime;
     /// use actix_rt::System;
     /// use futures_util::future::try_join_all;
@@ -175,20 +177,22 @@ impl System {
     ///
     /// let rest_operations = run_application();
     /// System::attach_to_tokio("actix-main-system", runtime, rest_operations);
+    /// # }
+    /// # #[cfg(not(feature = "tokio-executor"))] fn main() {} // Provie a blanket main function, so test will compile.
     /// ```
     pub fn attach_to_tokio<Fut, R>(
         name: impl Into<String>,
-        mut runtime: tokio::runtime::Runtime,
+        mut runtime: crate::executor::Runtime,
         rest_operations: Fut,
     ) -> R
     where
-        Fut: std::future::Future<Output = R>,
+        Fut: std::future::Future<Output = R> + 'static,
     {
         let actix_system_task = LocalSet::new();
         let sys = System::run_in_tokio(name.into(), &actix_system_task);
         actix_system_task.spawn_local(sys);
 
-        runtime.block_on(actix_system_task.run_until(rest_operations))
+        crate::executor::block_on_local(&mut runtime, &actix_system_task, rest_operations)
     }
 
     /// Get current running system.
--- a/actix-rt/src/system.rs
+++ b/actix-rt/src/system.rs
@@ -186,7 +186,7 @@ impl System {
         rest_operations: Fut,
     ) -> R
     where
-        Fut: std::future::Future<Output = R> + 'static,
+        Fut: std::future::Future<Output = R>,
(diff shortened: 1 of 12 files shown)


popzxc:
@actix/contributors does anyone want to review it?

actix-rt/src/system.rs
```diff
@@ -175,20 +177,22 @@ impl System {
...
     /// ```
     pub fn attach_to_tokio<Fut, R>(
         name: impl Into<String>,
-        mut runtime: tokio::runtime::Runtime,
+        mut runtime: crate::executor::Runtime,
         rest_operations: Fut,
     ) -> R
     where
-        Fut: std::future::Future<Output = R>,
+        Fut: std::future::Future<Output = R> + 'static,
```
Neopallium, line 189: Is adding ` + 'static` a breaking change to the API?  Is it required for both executors?

Neopallium (commented):
Other then adding ` + 'static` it looks good.  Might be early enough to add that bounds.

actix-rt/src/system.rs
popzxc: In fact it seems that neither of them required `'static`. Removed.

Neopallium (approved):
Looks good.

popzxc:
@robjtede Can you please review it?

popzxc:
@robjtede ? 🙃

Mcat12:
[Tokio 0.3](https://tokio.rs/blog/2020-10-tokio-0-3) was just released, along with a new [tokio-compat-02](https://crates.io/crates/tokio-compat-02) crate. Will that require another specialized feature, or can there be a more general way to use a different executor for workers which wouldn't require changes to support the new runtime? I can see a new #163-type issue being opened in the future around Tokio 0.2 and 0.3 compatibility.

JohnTitor (commented):
This is related to the above comment but I don't _feel_ like we should support `tokio-compat` (i.e. tokio v0.1) via `tokio-compat-executor` feature now, as the latest version of tokio is now v0.3 and the team is moving on v1.

And, I'm not sure if we should also support `tokio-compat-02` as migrating v0.2 -> v0.3 should be easier than v0.1 -> v0.2 AFAIK.

I'd like to take care of #197 and #203 atm, and this could be stale after resolving them. So I don't feel like we should review/merge this for now.

> can there be a more general way to use a different executor for workers which wouldn't require changes to support the new runtime?

Yeah, I imagine it'd be the best way here.

fakeshadow:
I like the idea of going runtime independent but we have too much tokio dep inside various modules. It would require a major rework far beyond `actix-rt` itself. So I personally think a good timing for that could be when `AsyncRead/Write` land in std and even after global executor land too.

fakeshadow:
I have a try in https://github.com/actix/actix-net/pull/208. Honest I don't know if this is the right thing so it would be helpful if someone can review this [test](https://github.com/fakeshadow/actix-net/blob/9a7a626f7bc251df723492a4e1da6d4bd641e9e4/actix-rt/tests/integration_tests.rs#L100)

robjtede:
I'm going to close this. I think there is agreement among the team that supporting Tokio v0.1 is not something we want to do, especially now that v1 is out of the door. Runtime independence is something I'd like to get to as well, eventually.

PR #207, opened 2020-10-28
remove RUNNING Q PENDING thread locals from actix-rt

### LINKED ISSUE #205: Contention in arbiter code
Disclaimer: I understand very little about how this works.

I'm using actix_web for my app and it works fabulously most of the time, thank you for working on it. My app is doing ~150krps from thousands of concurrent clients while burning half of a single 32 core Zen2 VM in GCP.

The issue I'm having is that sometimes my app would enter a state where all CPUs are burning mostly in scheduling code, while barely accepting any client connections and making health checks unhappy. Number of open file descriptors quickly blows well over a million and memory balloons over 32GB (mostly not RSS, but also not SLAB, so I'm quite bamboozled there too), when normally it's well under under 8GB.

Here's how an on-CPU flamegraph looks like (there's Kafka on the right, it's safe to ignore it):

image

At the top we're looking at mostly `try_read_output` doing most of the work:

image

Zoomed into the left pillar:

image

Zoomed into the right pillar:

image

One thing that helps here is not allowing clients to reuse connections, which keeps number of open files in check, but also cripples throughput. I have two services in a single app (client facing an metrics), and metrics is accessible just fine when client facing is suffering. Same happens when I replace client facing code with a service that just replies with 200 OK without doing anything.

Compare this to a flamegraph in a steady state (purple is `try_read_output`):

image

It's not exactly great with respect to spinlock contention in kernel networking code on the right (I suspect it's actix related, but it's not something I had time to confirm or worry about), but the problematic pillars are much narrower and they are not as exclusive to `try_read_output`, there are some other cpu consumers in them as well.

### LINKED ISSUE #206: Consider remove RUNNING, Q and PENDING thread locals from actix-rt
For now every `actix_rt::spawn` would go through these thread locals. 
This brings up several questions:

1. We check `RUNNING` and if it's false we box the future and push it to `Q`. This seems only useful when we want to spawn futures before tokio runtime setup but interanlly we can totally aviod that. Is it useful in user code? 

2. Whenever a future is spawned we push it's `JoinHandle` to `PENDING` and if `PENDING` goes large enough we would actively poll these handles and remove the resolved ones. This seems useful but what if user have a stream oriented endpoint that would spawn a long live future with every incoming request. Wouldn't that make `PENDING` grows very big and we would poll them all once with every new spawn?

3. `PENDING` is also used for `Arbiter::local_join` for clearing all pending futures by wait for them to resolve. This features is kind useful but I can't find anywhere we use it. I serached actix and actix-web repo and it's no where to be found.

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview
These thread locals are either outdated or are slowing down actix-rt when it's under heavy work load with long live streams. I have a feeling there are more issues related to it besides the issues mentioned.

This PR would break both actix-web and actix.

Closes #205 
Closes #206


actix-rt/src/lib.rs
```diff
@@ -1,6 +1,5 @@
 //! A runtime implementation that runs everything on the current thread.
-#![deny(rust_2018_idioms, warnings)]
```
robjtede, line 2: leave the idioms in, pls

fakeshadow: ok. would do

robjtede: sorry i meant just idioms, deny(warnings) is bad

fakeshadow: ok. I didn't think when adding it back. would remove it later

robjtede:
Did we ever get the bottom of the local_join situation. It was added so must have a valid use case. Is there an alternative in this PR?

fakeshadow:
I didn't find any usage of it in actix and actix-web.It could be used in user code so I would expect breaking.
 
Polling large amount of join handle at the same time is considerde harmful with the budget system so we should not doing that.
If we want to keep the feature we need a background future to recycle the join handles peroidicly or return join handle with spawns and users can collect the handles in use code.(Not all spawns can be collected though)

fakeshadow:
Re add `Arbiter::local_join`. `JoinHandle` of spawned tasks are pushed to `futures::stream::FuturesUnordered`. With every `actix_rt::spawn` or `Arbiter::send` we would also unconditionally poll `FuturesUnordered` once.

This would save some cost from constant polling all join handles. But in return it would also leak some join handles in the long run. Additional logic for polling multiple times could help on this issue but it's hard to tell how many polling is done for certain join handles and it would be very easy to overshoot above the coop buget.

actix-rt/src/runtime.rs
```diff
@@ -84,7 +88,7 @@ impl Runtime {
     /// complete execution by calling `block_on` or `run`.
     pub fn block_on<F>(&mut self, f: F) -> F::Output
     where
-        F: Future + 'static,
+        F: Future,
```
robjtede, line 91: Can we get a test case set up to prevent regression.

actix-rt/CHANGES.md
```diff
@@ -6,6 +6,19 @@
 
 * Add `System::attach_to_tokio` method. [#173]
 
+### Changed
+
+* Remove `'static` lifetime requirement for `Runtime::block_on` and `SystemRunner::block_on`. 
+  Remove `'static` lifetime requirement for `System::run`. 
+  Deprecate `Arbiter::is_running` and `Arbiter::local_join`.
```
robjtede, line 13: Local join no longer deprecated.

actix-rt/src/runtime.rs
fakeshadow: I have one test for this added in the tokio 0.3 pr. I would copy/paste it here.

fakeshadow:
Added test for non static block_on.
Also add an not very aggressive recycle function for join handles. 
The goal is to keep the totall poll time for every spwan below 128 times. I don't have evidence that this would limit the resource comsumption and it's based on a guess.

fakeshadow:
I still keen to remove the local_join function.

The reason is that it's in general is against the zero cost nature and force an extra cost on everyone no matter they use it or not. And the cost is higher when you have a high workload while being little when your app sits doing nothing.

The current setting means for anyone has more than 64 concurrent spawn tasks running for long time.(SSE, Ws or other background tasks) would pay for extra 64 - 128 polls on the executor with every new spawned tasks.

On the other hand the benefit is little to non. A function which is not used at all in any actix repo. (I doubt people use it in actix-web much either. As it's called on Arbiter which is mostly an actix feature. You have access to Arbiter in actix-web but it's not like there is feature you need it for.)

It's also worth to adding that actix-web's graceful shutdown server wait for every spawned http request finish and only drop the user spawned tasks without waiting on them. Which means user can collect join handles they want to wait for manually and wait on them.

That said my test from this PR has noticeable lower overhead than the current implementation so it could be an acceptable middle ground.

robjtede:
I'm not opposed to dropping the function but it was added by @Jonathas-Conceicao presumably for some use case. I'd really like to get clarity on what it was added for.

Maybe (a big maybe) it could be behind a crate feature flag.

robjtede (approved):

PR #214, opened 2020-11-24
fix clippy warnings

robjtede:
\+ make my spelling checker happy


fakeshadow (approved):
Maybe we should bump MSRV already. tokio 0.3 fail to compile on 1.42 and the MSRV is labled as 1.45

robjtede:
I don't want to bump MSRV just yet. I can think of a couple things that will need changes here for actix-web features/fixes.

PR #217, opened 2020-11-30
Add a MapRequest middleware for preprocessing requests

jplatte:
## PR Type

Feature

## PR Checklist

## Overview

This is not much more than a dump of some code we have in our code base, but that would benefit from being included in actix-service. For actix-web users, it allows the equivalent of

```rust
.wrap_fn(|req, srv| async move {
    let req = my_preprocess_fn(req).await?;
    srv.call(req).await
})
```

or

```rust
.wrap_fn(|req, srv| my_preprocess_fn(req).and_then(|req| srv.call(req)))
```

(both don't work because of lifetime issues)

by just writing

```rust
.wrap(MapRequest(my_preprocess_fn))
```

which even if the first two snippets worked, is clearer and shorter, and thus desireable if done a lot.

I could write some docs and tests, but I'm not even 100% certain the implementation is correct (specifically `poll_ready`), and it isn't as efficient as it could be (the `LocalBoxFuture` shouldn't be necessary). Also whether this is the right naming wasn't clear when discussing the feature on Gitter previously.

I'd be really grateful if somebody could pick this up.

CC @robjtede who has been part of one or two discussions around this feature

### DIFF
--- a/actix-service/src/lib.rs
+++ b/actix-service/src/lib.rs
@@ -19,6 +19,7 @@ mod map;
 mod map_config;
 mod map_err;
 mod map_init_err;
+mod map_request;
 mod pipeline;
 mod then;
 mod transform;
@@ -28,6 +29,7 @@ pub use self::apply::{apply_fn, apply_fn_factory};
 pub use self::apply_cfg::{apply_cfg, apply_cfg_factory};
 pub use self::fn_service::{fn_factory, fn_factory_with_config, fn_service};
 pub use self::map_config::{map_config, unit_config};
+pub use self::map_request::{MapRequest, MapRequestMiddleware};
 pub use self::pipeline::{pipeline, pipeline_factory, Pipeline, PipelineFactory};
 pub use self::transform::{apply, Transform};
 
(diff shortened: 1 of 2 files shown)


fakeshadow:
I feel this is a middleware that can live in `actix_web::middlewares`.
The poll_ready is correct. As poll methods are always sync and exclusive.

That said why not just use `warp_fn(my_preprocess_fn)`. MapRequest is not much of a reduce of code from an extra line of `srv.call(req).await`.

jplatte:
> That said why not just use `warp_fn(my_preprocess_fn)`. MapRequest is not much of a reduce of code from an extra line of `srv.call(req).await`.

I think the PR description is pretty clear about that, no? If you mean why not move `srv.call(req).await` into the fn instead of using a closure, that requires quite some digging to figure out the correct signature, since all examples use closures. Even if the right signature was documented, it would still be much more complicated and require more things to be imported.

robjtede:
closing due to inactivity, we'll see about picking this up again if someone asks

PR #222, opened 2020-12-04
fix "Can not register server socket" The parameter is incorrect. on windows

### LINKED ISSUE #221: Can not register server socket The parameter is incorrect. (os error 87)
This is a problem that appear to be related to https://github.com/actix/actix-web/issues/905.

I have this with the latest version of actix-web 3.3.2 on windows 10.

danylaporte:
## PR Type

Bug Fix

## PR Checklist

## Overview

This should fix another place where this issue have occured.

Closes #221

### DIFF
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -266,12 +266,7 @@ impl Accept {
         for (token, info) in self.sockets.iter_mut() {
             if let Some(inst) = info.timeout.take() {
                 if now > inst {
-                    if let Err(err) = self.poll.register(
-                        &info.sock,
-                        mio::Token(token + DELTA),
-                        mio::Ready::readable(),
-                        mio::PollOpt::edge(),
-                    ) {
+                    if let Err(err) = register(&self.poll, token, info) {
                         error!("Can not register server socket {}", err);
                     } else {
                         info!("Resume accepting connections on {}", info.addr);
@@ -298,7 +293,7 @@ impl Accept {
                     }
                     Command::Resume => {
                         for (token, info) in self.sockets.iter() {
-                            if let Err(err) = self.register(token, info) {
+                            if let Err(err) = register(&self.poll, token, info) {
                                 error!("Can not resume socket accept process: {}", err);
                             } else {
                                 info!(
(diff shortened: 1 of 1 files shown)


fakeshadow:
I don't unstand. What is the difference?

danylaporte:
2 more lines have been added to the `register` function that call `reregister` on the poll if the first `register` call failed.

robjtede:
I see the change. What's the case that causes it to hit this error?

danylaporte:
@robjtede I don't really know. The codebase is very big and private and I cannot produce a simplified code that reproduces the error. The error occurs some of the time, not always. When I hit the error, actix-web is restarting automatically.

robjtede:
Ok, just add a changelog entry and we can get this in.

fakeshadow:
The thing is if a socket has a timeout it's already deregistered previously. It's either fail to deregister or you should never call register when recovery from backpress.

robjtede:
fakeshadow makes a good point, it would be extraordinarily useful to get something documented instead of blindly patching something we think shouldn't be possible.

I understand that it's private code and minimal reproductions can be very hard to create in these situations. There's 2 error logs near here: https://github.com/actix/actix-net/blob/c02c4603bbb5e108d6d1d2effc5f67b69e35cd8c/actix-server/src/accept.rs#L420 (the only place which starts the timeout process that triggers process_timer) which might shed light on what went wrong here.

danylaporte:
Sometimes, I have the following in my log file:

`Error accepting connection: Either the application has not called WSAStartup, or WSAStartup failed. (os error 10093)`

Probably related to the deregister.

robjtede:
We've spotted something within the codepath that could be the cause of this. Can you try the patch in #223 @danylaporte?

danylaporte:
@robjtede I tried the #223 pull request and I still can reproduce the issue. I did not have any log in my file but actix-web restart. I will continue to investigate and put more log for debugging the issue.

danylaporte:
So frustrating that when I add logging code, I cannot reproduce the issue at all.

danylaporte:
The issue was caused by an invalid unsafe code not related to actix.

PR #224, opened 2020-12-12
Refactor LocalWaker

botika:
~Simplify LocalWaker register method with `replace` mutable pointer method~

Refactor `LocalWaker`: `is_registered` has no real use because the only use of the registry is to wake up the task it refers to. But `wake` does not need the registration status, the behavior is the same whether a task is registered or not. So it has been removed in favor of `wake`. Without this method the use of `UnsafeCell` is not necessary so I have proposed a later refactor from `UnsafeCell` to `Cell`.

### DIFF
--- a/actix-server/src/signals.rs
+++ b/actix-server/src/signals.rs
@@ -2,9 +2,6 @@ use std::future::Future;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
-#[cfg(not(unix))]
-use std::io;
-
 use futures_util::future::lazy;
 
 use crate::server::Server;
@@ -26,7 +23,7 @@ pub(crate) enum Signal {
 pub(crate) struct Signals {
     srv: Server,
     #[cfg(not(unix))]
-    stream: Pin<Box<dyn Future<Output = io::Result<()>>>>,
+    stream: Pin<Box<dyn Future<Output = std::io::Result<()>>>>,
     #[cfg(unix)]
     streams: Vec<(Signal, actix_rt::signal::unix::Signal)>,
 }
--- a/actix-server/src/signals.rs
+++ b/actix-server/src/signals.rs
@@ -1,7 +1,9 @@
 use std::future::Future;
-use std::io;
 use std::pin::Pin;
 use std::task::{Context, Poll};
+
+#[cfg(not(unix))]
+use std::io;
 
 use futures_util::future::lazy;
 
@@ -30,7 +32,7 @@ pub(crate) struct Signals {
 }
 
 impl Signals {
-    pub(crate) fn start(srv: Server) -> io::Result<()> {
+    pub(crate) fn start(srv: Server) {
         actix_rt::spawn(lazy(|_| {
             #[cfg(not(unix))]
             {
@@ -66,8 +68,6 @@ impl Signals {
                 actix_rt::spawn(Signals { srv, streams })
             }
         }));
-
-        Ok(())
     }
 }
 
--- a/actix-utils/src/task.rs
+++ b/actix-utils/src/task.rs
@@ -36,7 +36,7 @@ impl LocalWaker {
     }
 
     #[deprecated(
-        since = "2.x.x",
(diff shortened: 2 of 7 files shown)


botika:
It seems there are clippy bugs in the current version. Now I upload the fix.

actix-server/src/signals.rs
```diff
@@ -1,8 +1,10 @@
 use std::future::Future;
-use std::io;
 use std::pin::Pin;
 use std::task::{Context, Poll};
 
+#[cfg(not(unix))]
+use std::io;
```
robjtede, line 6: inline this since it's only used in one place under the cfg attr

actix-utils/src/task.rs
```diff
@@ -45,12 +45,7 @@ impl LocalWaker {
     ///
     /// Returns `true` if waker was registered before.
     pub fn register(&self, waker: &Waker) -> bool {
-        unsafe {
-            let w = self.waker.get();
-            let is_registered = (*w).is_some();
-            *w = Some(waker.clone());
-            is_registered
-        }
+        unsafe { self.waker.get().replace(Some(waker.clone())).is_some() }
```
robjtede, line 48: Binding names are good.

```rust
let w = self.waker.get();
let last_waker = w.replace(Some(waker.clone()));
// indicates was registered
last_waker.is_some()
```

robjtede (changes requested):

fakeshadow:
We could just use a Cell though. The counter's waker is rarely used and when it's used the system could already be bound to CPU.

botika:
> We could just use a Cell though. The counter's waker is rarely used and when it's used the system could already be bound to CPU.

`Waker` not implement `Copy`

fakeshadow:
https://github.com/fakeshadow/actix-net/blob/2d35038600cc74b527f6e84e5dcf50e27ed48c48/actix-rt/src/net.rs#L106
This is the counter I use in a fork.

botika:
https://doc.rust-lang.org/std/cell/struct.Cell.html#impl-1
Well, I guess you won't use the get method at all.

fakeshadow:
Yea we don't use that. `is_registered` is an API I can't find usage internally. Even the register API just replace and check the old one as option

botika:
The uses of LocalWaker `take` in `actix-web`:
- https://github.com/actix/actix-web/blob/b5812b15f07271850367b085e4933a0989dd84f9/actix-http/src/h1/payload.rs#L185
- https://github.com/actix/actix-web/blob/b5812b15f07271850367b085e4933a0989dd84f9/actix-multipart/src/server.rs#L728-L730
Can substitute by `wake`. I don't think I left any.

I `is_registered` is not used in `actix-web`

robjtede:
I wouldn't create breaking change without good reason at this point. Keep this as a refactor.

fakeshadow:
I think it's OK to remove `is_registered`. There is no instance of using it in any of actix repo and I doubt anyone using it by importing actix-utils for a waker type.

Besides that `LocalWaker` doesn't expose an API for comparing the waker so you know it's registered but you don't know if you still have the same context by comparing the waker.(Which could have false positive but I'm not sure).
I feel it's safer to always replace the waker than check it.

robjtede:
Changing public API is always a breaking change. Please keep it the same for this PR. Adding deprecation attribute to `is_registered` with a note about its lack of usefulness would be acceptable.

botika:
> Changing public API is always a breaking change. Please keep it the same for this PR. Adding deprecation attribute to `is_registered` with a note about its lack of usefulness would be acceptable.

Totally agree. Sorry now I put this methods deprecated. I do not know what I was thinking of.

actix-server/src/signals.rs
robjtede: this has come back

botika: Sorry :parachute:  Now it's fixed

robjtede (approved):

fakeshadow (approved):

PR #228, opened 2020-12-15
Standardize bytes dependencies to version 0.5.6

botika:
Use last compatible version of `bytes` with `tokio@0.2`


robjtede:
this kind of change is unnecessary due to the semver system

PR #248, opened 2021-01-18
change default name resolver and allow custom resolvers

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove `trust-dns` dependencies. 
Use `tokio::net::lookup_host` for simple dns lookup to improve compile time and reduce deps.

Make `Resolver` an enum type that accept custom `Resolve` trait object for custom resolver implementation.

Remove `either` and `futures-util` dep and simplify the connector and service logic.

Rework `connect` mod and remove duplicated iterator types. Reduce nested `Either` type and simplify Addrs to an enum.

export `webpki_roots::TLS_SERVER_ROOTS` in `connect` mod and remove it from `accept` mod. This is a type used only on client

Add more `connect` mod doc.

Remove commented code from `actix-rt`

### DIFF
--- a/actix-tls/src/connect/resolve.rs
+++ b/actix-tls/src/connect/resolve.rs
@@ -1,184 +1,123 @@
-use std::future::Future;
-use std::marker::PhantomData;
-use std::net::SocketAddr;
-use std::pin::Pin;
-use std::task::{Context, Poll};
+use std::{
+    net::{SocketAddr, ToSocketAddrs},
+    rc::Rc,
+    task::Poll,
+};
 
 use actix_service::{Service, ServiceFactory};
-use futures_util::future::{ok, Either, Ready};
+use futures_core::future::LocalBoxFuture;
 use log::trace;
-use trust_dns_resolver::TokioAsyncResolver as AsyncResolver;
-use trust_dns_resolver::{error::ResolveError, lookup_ip::LookupIp};
 
 use super::connect::{Address, Connect};
 use super::error::ConnectError;
-use super::get_default_resolver;
 
 /// DNS Resolver Service factory
-pub struct ResolverFactory<T> {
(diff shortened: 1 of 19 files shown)


robjtede:
trust dns is useful for some set ups, i'd prefer to have it as an optional feature, off by default

fakeshadow:
I'm afraid it would furtuer complicate the feature flag. We have to make feature flags for actix-tls, actix-http and awc.

I believe it would be easier to include a detail example on how to implement trust-dns as custom resolver and pass it to the constructor of awc. It can be done in small amount of code.

robjtede:
abstracting over resolvers sounds fine as a solution, it can be rolled into the awc rewrite

robjtede (approved):
LGTM

actix-tls/src/connect/resolve.rs
```diff
@@ -1,184 +1,225 @@
...
+/// let resolver = Resolver::new_custom(resolver);
+///
+/// // pass custom resolver to connector builder.
+/// // connector would then be usable as a service or awc's connector.
+/// let connector = actix_tls::connect::new_connector::<&str>(resolver.clone());
+///
+/// // resolver can be passed to connector factory where returned service factory
+/// // can be used to construct new connector services.
+/// let factory = actix_tls::connect::new_connector_factory::<&str>(resolver);
+///```
```
pickfire, line 106: I think you missed a space here.

pickfire:
Can the custom dns resolver request an acme cert from lets encrypt when upon client request? I did another project in tokio but I am wondering if it could be done in actix.

fakeshadow:
I'm not sure. The dns resolver in actix-tls would take a String/&'static str/http::Uri type and return a usable SocketAddr and that's basically all it can do.

PR #249, opened 2021-01-22
expose create_tcp_listener function. do not enable reuse_addr on windows

fakeshadow:
## PR Type

Other

## PR Checklist

## Overview

Expose `actix_server::builder::create_tcp_listener` function. With it actix-web can detch it's own tcplistener constructor function.
Do not set reuse_addr to true on windows platform. https://github.com/actix/actix-web/issues/1913

### DIFF
--- a/actix-server/CHANGES.md
+++ b/actix-server/CHANGES.md
@@ -2,8 +2,10 @@
 
 ## Unreleased - 2021-xx-xx
 * Hidden `ServerBuilder::start` method has been removed. Use `ServerBuilder::run`. [#246]
+* Expose `builder::create_tcp_listener`. Don't set reuseaddr on windows. [#249]
 
 [#246]: https://github.com/actix/actix-net/pull/246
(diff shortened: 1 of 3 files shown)


robjtede:
I'm not a fan of more windows specific code. Exposed function should provide option or builder.

fakeshadow:
The goal is to change default behavior on listener constructing. I can't think a way other than cfg.
The exposed function is for actix-web and some tests for a reusable easy builder. It's not meant for user.

I'm closing this for now as I don't know how to change it to not be cfg specific.

robjtede:
Maybe I was too hasty in denying this. After doing some more reading seems like a reasonable Windows-specific difference.

As I understand it now, SO_REUSEADDR on UNIX only binds port if socket is in TIME_WAIT state (or otherwise not listening) but on Windows it will always bind. Eg. doing two `cargo run` on Windows will succeeed? That's not great.

robjtede:
OTOH, this would add mio to our public api so maybe not.

PR #252, opened 2021-01-28
fix actix-tls build

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

fix openssl and rustls build

### DIFF
--- a/actix-rt/tests/integration_tests.rs
+++ b/actix-rt/tests/integration_tests.rs
@@ -172,17 +172,3 @@ fn attack_to_tokio() {
 
     assert_eq!(res, 2);
 }
-
-#[tokio::test]
-async fn attack_to_tokio_macro() {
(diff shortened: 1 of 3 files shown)


fakeshadow:
Didn't notice there is unworking test in actix-rt. I removed the `attach_tokio_macro` test because a tokio handler can only be used to spawn task. It can not call `block_on` as that would possibly make nested `block_on` happen.

The possible solutation would be generic attach_to_tokio over both `&Runtime` and `Handler` type with a trait and call `spawn` in it. Which would make the return type a `JoinHandle<Future::Output>`. And then the caller can block_on it manually in main.

robjtede:
> Or what I prefer just remove APIs about running actix system in existing tokio runtime all together. 

I'm leaning towards this also.

PR #253, opened 2021-01-28
remove tokio runners

robjtede:
## PR Type
Refactor

## PR Checklist

## Overview
- Remove `run_in_tokio` and `attach_to_tokio`.
- Remove `AsyncSystemRunner`.
- Return `JoinHandle` from `actix_rt::spawn`.
	- Remove old `Arbiter::spawn`. Implementation is now inlined into `actix_rt::spawn`.
- Rename `Arbiter::{send => spawn}` and `Arbiter::{exec_fn => spawn_fn}`.
- Remove `Arbiter::exec`.
- Remove deprecated `Arbiter::local_join`.

### DIFF
--- a/actix-rt/CHANGES.md
+++ b/actix-rt/CHANGES.md
@@ -2,6 +2,16 @@
 
 ## Unreleased - 2021-xx-xx
 
+* Remove `run_in_tokio` and `attach_to_tokio`. [#253]
+* Remove `AsyncSystemRunner`. [#253]
+* Return `JoinHandle` from `actix_rt::spawn`. [#253]
+* Remove old `Arbiter::spawn`. Implementation is now inlined into `actix_rt::spawn`. [#253]
+* Rename `Arbiter::{send => spawn}` and `Arbiter::{exec_fn => spawn_fn}`. [#253]
+* Remove `Arbiter::exec`. [#253]
+* Remove deprecated `Arbiter::local_join`. [#253]
+
+[#253]: https://github.com/actix/actix-net/pull/253
+
 
 ## 2.0.0-beta.2 - 2021-01-09
 * Add `task` mod with re-export of `tokio::task::{spawn_blocking, yield_now, JoinHandle}` [#245]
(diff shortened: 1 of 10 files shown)


nooberfsh:
Why it is removed? 
Many people use this to run actix inside tokio runtime.  https://github.com/actix/actix-web/issues/1283

fakeshadow:
It's removed for misleading and not working properly.

1. actix always runs in tokio. It never support runtime other than tokio from what I know.
2. The said API only attach `SystemArbiter` to an existing tokio runtime. Every new `Arbiter`(include actix-web worker) would still spawn it's tokio runtime regardless. Even when you limit your actix-web worker count to 1 that worker would still run on it's own tokio runtime. It has nothing to do with the runtime you pass to `run_in_tokio` unless you explicitly send boxed future to `SystemArbiter`.
3. There is nothing wrong with running multiple runtimes in an app. `actix` and `actix-web` are examples on this where you can have more than dozens of runtime working on their own at the same time. Therefore if you need another tokio runtime you can spawn it along side your `actix-rt`. There is no observable overhead on doing this.

Overdash:
> 3. There is nothing wrong with running multiple runtimes in an app. `actix` and `actix-web` are examples on this where you can have more than dozens of runtime working on their own at the same time. Therefore if you need another tokio runtime you can spawn it along side your `actix-rt`. There is no observable overhead on doing this.

Do you have an example of doing this (spawning another tokio runtime along side an actix system)?

From what I see, instances like `actix-web` HttpServer creation are blocking. So in this case it's not possible/difficult to spawn a usable tokio Runtime that can be used by all the endpoints (for example). I could be missing something here though

fakeshadow:
> From what I see, instances like `actix-web` HttpServer creation are blocking. So in this case it's not possible/difficult to spawn a usable tokio Runtime that can be used by all the endpoints (for example). I could be missing something here though

Please check point two. You `endpoints` run on workers that are in their own runtime which is not in your main or have have anything to do with the runtime you set up in mian. Except the condition described by it.

PR #265, opened 2021-02-03
add blocking thread customize

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Make blocking thread pool count changable at per worker level.

### DIFF
--- a/actix-server/src/builder.rs
+++ b/actix-server/src/builder.rs
@@ -78,20 +78,19 @@ impl ServerBuilder {
         self
     }
 
-    /// Set number of max blocking threads for running blocking task on thread pool.
+    /// Set max number of threads for each worker's blocking task thread pool.
     ///
-    /// * This thread pool is per worker and `NOT` shared across workers.
+    /// One thread pool is set up **per worker**; not shared across workers.
     ///
     /// # Examples:
-    /// ```rust
+    /// ```
     /// # use actix_server::ServerBuilder;
     /// let builder = ServerBuilder::new()
     ///     .workers(4) // server has 4 worker thread.
     ///     .worker_max_blocking_threads(4); // every worker has 4 max blocking threads.
     /// ```
     ///
-    /// See [tokio::runtime::Builder::max_blocking_threads](tokio::runtime::Builder::max_blocking_threads)
-    /// for behavior reference.
+    /// See [tokio::runtime::Builder::max_blocking_threads] for behavior reference.
     pub fn worker_max_blocking_threads(mut self, num: usize) -> Self {
         self.worker_config.max_blocking_threads(num);
         self
--- a/actix-server/src/builder.rs
+++ b/actix-server/src/builder.rs
@@ -19,7 +19,7 @@ use crate::signals::{Signal, Signals};
 use crate::socket::{MioListener, StdSocketAddr, StdTcpListener, ToSocketAddrs};
 use crate::socket::{MioTcpListener, MioTcpSocket};
 use crate::waker_queue::{WakerInterest, WakerQueue};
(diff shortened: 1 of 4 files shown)


actix-server/src/builder.rs
```diff
@@ -78,6 +78,25 @@ impl ServerBuilder {
...
+    /// # Examples:
+    /// ```rust
+    /// # use actix_server::ServerBuilder;
+    /// let builder = ServerBuilder::new()
+    ///     .workers(4) // server has 4 worker thread.
+    ///     .worker_max_blocking_threads(4); // every worker has 4 max blocking threads.
+    /// ```
+    ///
+    /// See [tokio::runtime::Builder::max_blocking_threads](tokio::runtime::Builder::max_blocking_threads)
+    /// for behavior reference.
```
robjtede, line 94:
```suggestion
    /// See [tokio::runtime::Builder::max_blocking_threads] for behavior reference.
```

actix-server/src/builder.rs
```diff
@@ -78,6 +78,25 @@ impl ServerBuilder {
         self
     }
 
+    /// Set number of max blocking threads for running blocking task on thread pool.
+    ///
+    /// * This thread pool is per worker and `NOT` shared across workers.
+    ///
+    /// # Examples:
+    /// ```rust
```
robjtede, line 86:
```suggestion
    /// ```
```

actix-server/src/builder.rs
```diff
@@ -78,6 +78,25 @@ impl ServerBuilder {
         self
     }
 
+    /// Set number of max blocking threads for running blocking task on thread pool.
```
robjtede, line 81:
```suggestion
    /// Set max number of threads for each worker's blocking task thread pool.
```

actix-server/src/builder.rs
```diff
@@ -78,6 +78,25 @@ impl ServerBuilder {
...
+    ///
+    /// * This thread pool is per worker and `NOT` shared across workers.
```
robjtede, line 83:
```suggestion
    /// One thread pool is set up **per worker**; not shared across workers.
```

actix-server/src/worker.rs
```diff
@@ -159,26 +159,61 @@ enum WorkerServiceStatus {
     Stopped,
 }
 
+// config for worker behavior passed from server builder.
```
robjtede, line 162:
```suggestion
/// Config for worker behavior passed down from server builder.
```

actix-server/src/worker.rs
```diff
@@ -159,26 +159,61 @@ enum WorkerServiceStatus {
...
+    shutdown_timeout: Duration,
+    max_blocking_threads: usize,
+}
+
+impl Default for ServerWorkerConfig {
+    fn default() -> Self {
+        Self {
+            shutdown_timeout: Duration::from_secs(30),
+            // default max_blocking_threads is the same as tokio's default
+            max_blocking_threads: 512,
```
robjtede, line 174: I think it should be a bit lower to account for common case of multi-worker so as to not potentially blow up thread count. 512 / 4 sounds good. So 128.

robjtede (changes requested):
looks good, just some doc nits

actix-server/src/worker.rs
fakeshadow: What about divide 512 with the core count num_cpus returns.

robjtede: good idea, i guess we have that as a dep anyway

robjtede (approved):

robjtede:
Somewhat arguable it should be `512 / #workers` but this is fine.

PR #275, opened 2021-02-06
add System::try_current

robjtede:
## PR Type
Feature

## PR Checklist

## Overview
Add `System::try_current`. Use case is to be able to stop Systems whether they exist or not without panicking.

Tokio uses a Result for their `Handle::try_current` method but it seems unecessary to make a new type just for it when you can just `.ok_or(...)` at the call site.


fakeshadow (approved):
LGTM

PR #286, opened 2021-02-24
refactor dispatcher / add Receiver::recv 

robjtede:
## PR Type

Refactor

## PR Checklist
- [x] Added tests.

## Overview
~~Reduce maintenance burden.~~ Keep mpsc public since it is exposed in public API of Dispatcher.

Refactor a few things: see changelog.


fakeshadow:
LGTM. 

I don't think async method is needed as it would always be used in a leaf future but it's Ok to have it.
I also find poll_fn kinda hacky but it's a nice util to have. I feel it would be better to construct a typed future for poll_fn like usage where the purpose is more clear reflexed by the type's name and code more readable.

fakeshadow (approved):

PR #295, opened 2021-03-24
Add native tls support for actix_tls::connect module

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

native tls was lacking on client connect module while impled in server accept module.
This PR add support for it in client connect module and can be enabled with `native-tls` feature.

No test was added for this feature as current  tests does not actually test any tls connectors(no matter openssl, rustls or native-tls). So there is no point adding test for it until this issue sorted out.

Some clippy warning fix for nightly has also been worked into this PR.


actix-tls/src/connect/ssl/mod.rs
```diff
@@ -5,3 +5,6 @@ pub mod openssl;
 
 #[cfg(feature = "rustls")]
 pub mod rustls;
+
+#[cfg(feature = "native-tls")]
+pub mod nativetls;
```
robjtede, line 10: nit but the mod name should match the crate name like `native_tls`

fakeshadow: yea. I was match the naming with accept module. Should this be a break change PR and re-name that too?

robjtede: sure

robjtede (approved):

robjtede:
theres a conflict i can't solve in the UI apparently

PR #296, opened 2021-03-24
fix rustls panic when generating dns name from ip

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

- [x] Add changelog entry.

## Overview

Return error instead of panic when rustls failed to generate Dns name ref from a given host(e.g. A host with ip address).

Closes https://github.com/actix/actix-web/issues/2100


robjtede (approved):
great! 👍🏻

can you add changelog entry

robjtede:
merge at your leisure

PR #310, opened 2021-04-02
Refactor WorkerState::Shutdown

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Remove the second sleep timer used by WorkerState::Shutdown variant.
This would make the shutdown timeout logic simpler with max of one second off of timeout accuracy.(Default 30s may become 31s worst case).

### DIFF
--- a/actix-server/src/worker.rs
+++ b/actix-server/src/worker.rs
@@ -1,17 +1,27 @@
-use std::future::Future;
-use std::pin::Pin;
-use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::Arc;
-use std::task::{Context, Poll};
-use std::time::Duration;
+use std::{
+    future::Future,
+    mem,
+    pin::Pin,
+    sync::{
+        atomic::{AtomicBool, AtomicUsize, Ordering},
+        Arc,
+    },
+    task::{Context, Poll},
+    time::Duration,
+};
 
-use actix_rt::time::{sleep, Sleep};
-use actix_rt::{spawn, Arbiter};
+use actix_rt::{
+    spawn,
+    time::{sleep, Instant, Sleep},
+    Arbiter,
+};
 use actix_utils::counter::Counter;
 use futures_core::{future::LocalBoxFuture, ready};
 use log::{error, info, trace};
-use tokio::sync::mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender};
-use tokio::sync::oneshot;
+use tokio::sync::{
+    mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
+    oneshot,
+};
 
 use crate::service::{BoxedServerService, InternalServiceFactory};
 use crate::socket::MioStream;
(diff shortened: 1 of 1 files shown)


actix-server/src/worker.rs
```diff
@@ -211,12 +221,12 @@ impl ServerWorker {
             let mut wrk = MAX_CONNS_COUNTER.with(move |conns| ServerWorker {
                 rx,
                 rx2,
+                services: Default::default(),
```
robjtede, line 224:
```suggestion
                services: Vec::new(),
```

actix-server/src/worker.rs
```diff
@@ -337,61 +347,69 @@ enum WorkerState {
...
-    Shutdown(
-        Pin<Box<Sleep>>,
-        Pin<Box<Sleep>>,
-        Option<oneshot::Sender<bool>>,
-    ),
+    // Shutdown keep states necessary for server shutdown:
+    // Sleep for interval check the shutdown progress.
+    // Instant for the start time of shutdown.
+    // Sender for send back the shutdown outcome(force/grace) to StopCommand caller.
+    Shutdown(Pin<Box<Sleep>>, Instant, oneshot::Sender<bool>),
```
robjtede, line 354: if theres a refactor of this, we should use named fields

actix-server/src/worker.rs
```diff
@@ -403,60 +421,61 @@ impl Future for ServerWorker {
...
-                let num = num_connections();
-                if num == 0 {
-                    let _ = tx.take().unwrap().send(true);
+            WorkerState::Shutdown(ref mut timer, ref start_from, _) => {
+                // Wait for 1 second.
+                ready!(timer.as_mut().poll(cx));
+
+                if num_connections() == 0 {
+                    // Graceful shutdown.
+                    if let WorkerState::Shutdown(_, _, sender) = mem::take(&mut this.state) {
```
robjtede, line 438: take sender state from outer match ?

fakeshadow: This is a perf trick to only copy the address when graceful shutdown happen.
Other variant only need to borrow the state rather than take it.

robjtede: It's an extra branch on an irrefutable pattern; the only difference is that this will set state to unavailable too. What's up with replacing the if-let with `sender.send(true)`

robjtede: It's an extra branch. The only difference is that mem::take will also set state back to unavailable but that doesn't change anything. What's up with `sender.send(true)`.

fakeshadow: This branch only happens when server goes into graceful shutdown so extra branch is not important at all. 

The `mem::take` is to take ownership of `Sender`. oneshot sender take self when calling send API. 
It was an Option<Sender> for the same reason of taking the ownership. But since Sender here is a tokio type so Rust have problem to optimize away the Option wrapper and it can take 8 bytes extra size.

robjtede: oneshot sender takes self... right

robjtede (approved):

PR #318, opened 2021-04-05
Fix bug where worker service restart could skip failing services and not being able to restart multiple services

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

When worker restarting multiple services that return error from `Service::poll_ready` the current logic would be throwing away all error except the last one in iteration. As previous service states are set to `WorkerServiceStatus::Failed` which would skip future readiness check and render they skip `poll_ready` check from then on.

When multiple services are started with `ServerBuilder::configure` API the factory service future would return a vec of `(Token, BoxedServerService)` type. And the restart logic would always take the first element from the vec. Making there is no way to restart any service other than the first one registered to `configure`.

This PR fixes these two bug with properly handle every error returned with `Service::poll_ready` and not skip to next untill the last one is resolved. After the service restart the `Token` of restarting service would be used to get the correct element in returned vector instead of only using the first one.

### DIFF
--- a/actix-server/src/worker.rs
+++ b/actix-server/src/worker.rs
@@ -404,16 +404,19 @@ impl Future for ServerWorker {
                 let factory_id = restart.factory_id;
                 let token = restart.token;
 
-                let mut item = ready!(restart.fut.as_mut().poll(cx)).unwrap_or_else(|_| {
-                    panic!(
-                        "Can not restart {:?} service",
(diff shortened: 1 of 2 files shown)


fakeshadow:
Added a test that would fail on current master branch but work with this PR.

robjtede (approved):

PR #319, opened 2021-04-06
Add pause flag for Accept

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

It's very hard to track the exact state of `Accept` and one more flag would help to record the information needed.

Use a u8 to tracking the state of backpressure and pause. This would prevent backpressure triggers when server in pause state, multiple call to pause, double deregister when server proceed to stop in pause state.

### DIFF
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -166,17 +166,14 @@ impl Accept {
         loop {
             if let Err(e) = self.poll.poll(&mut events, None) {
                 match e.kind() {
-                    std::io::ErrorKind::Interrupted => {
-                        continue;
-                    }
-                    _ => {
-                        panic!("Poll error: {}", e);
-                    }
+                    std::io::ErrorKind::Interrupted => continue,
+                    _ => panic!("Poll error: {}", e),
                 }
             }
 
             for event in events.iter() {
                 let token = event.token();
+
                 match token {
                     // This is a loop because interests for command from previous version was
                     // a loop that would try to drain the command channel. It's yet unknown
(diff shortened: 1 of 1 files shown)


fakeshadow:
Test example.

```rust
use std::{env, io};

use actix_rt::net::TcpStream;
use actix_server::Server;
use actix_service::{fn_factory, fn_service};
use tokio::io::AsyncWriteExt;

#[actix_rt::main]
async fn main() -> io::Result<()> {
    env::set_var("RUST_LOG", "actix=error,basic=error");
    env_logger::init();

    let addr = ("127.0.0.1", 8080);

    let server = Server::build()
        .workers(2)
        .maxconn(128)
        .bind("echo", addr, move || {
            fn_factory(|| async {
                Ok::<_, ()>(fn_service(|mut stream: TcpStream| async move {
                    let buf = b"HTTP/1.1 200 OK\r\n\
                    content-length: 12\r\n\
                    connection: close\r\n\
                    date: Thu, 01 Jan 1970 12:34:56 UTC\r\n\
                    \r\n\
                    Hello World!\
                    ";

                    stream.write(buf).await.unwrap();
                    Ok::<_, ()>(())
                }))
            })
        })?
        .run();

    let handle = server.clone();

    std::thread::spawn(move || {
        loop {
            std::thread::sleep(std::time::Duration::from_secs(1));
            let _ = handle.pause();
            std::thread::sleep(std::time::Duration::from_secs(1));
            let _ = handle.resume();
        }
    });

    server.await
}
```

fakeshadow:
Alternative flag has been added to #343

Issue #216, opened 2020-11-29
Limit of max_connections is not enforced on backlog


bobrik:
Here's an example service with `max_connections` set to 2 and `backlog` of 16 (`actix-web` v3.3.0):

```rust
const MAX_CONNECTIONS: usize = 2;
const BACKLOG: i32 = 16;
const WORKERS: usize = 1;

const ADDR: &str = "127.0.0.1:8888";

#[actix_web::main]
async fn main() -> std::io::Result<()> {
    simplelog::CombinedLogger::init(vec![simplelog::TermLogger::new(
        simplelog::LevelFilter::Info,
        simplelog::Config::default(),
        simplelog::TerminalMode::Stderr,
    )])
    .expect("logging configuration failed");

    log::info!(
        "Running server with {} workers / {} max connections / {} listen backlog on {}",
        WORKERS,
        MAX_CONNECTIONS,
        BACKLOG,
        ADDR
    );

    actix_web::HttpServer::new(move || {
        actix_web::App::new().route("/", actix_web::web::get().to(greet))
    })
    .on_connect(move |_, _| {
        log::info!(
            "number of open fds = {}",
            procfs::process::Process::myself()
                .expect("creating procfs reader failed")
                .fd_count()
                .expect("getting fd count failed")
        )
    })
    .max_connections(MAX_CONNECTIONS)
    .backlog(BACKLOG)
    .workers(WORKERS)
    .bind(ADDR)?
    .run()
    .await
}

async fn greet() -> impl actix_web::Responder {
    "howdy!\n"
}
```

One can run it:

```
$ cargo run
```

And then use the following go program to create one connection every 100ms without any requests:

```go
package main

import (
	"log"
	"net"
	"time"
)

func main() {
	i := 0
	for {
		i++

		_, err := net.Dial("tcp", "127.0.0.1:8888")
		if err != nil {
			log.Fatalf("Connection #%d failed: %s", i, err)
		}

		log.Printf("Connnection #%d succeded", i)

		time.Sleep(time.Millisecond * 100)
	}
}
```

The go program will happily churn until the end of ulimit.

What I expect from Actix is to not have more than 2 (limit) + 16 (backlog) connections open at any point, which corresponds to 22 open file descriptors (let's say under 30 for simplicity). What I see is that number gets a lot higher:

```
01:32:00 [INFO] Running server with 1 workers / 2 max connections / 16 listen backlog on 127.0.0.1:8888
01:32:00 [INFO] Starting 1 workers
[...]

PR #322, opened 2021-04-08
do no drain backlog on backpressure

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Do not drain backlog when server enter backpressure.

### DIFF
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -407,7 +407,7 @@ impl Accept {
     }
 
     fn accept(&mut self, sockets: &mut Slab<ServerSocketInfo>, token: usize) {
-        loop {
+        while !self.backpressure {
             let info = sockets
                 .get_mut(token)
                 .expect("ServerSocketInfo is removed from Slab");


bobrik:
This helps, but it doesn't solve #216 completely.

fakeshadow:
Well the backpressure happens when holding a connection so either we force push it to worker or drop it.

It also worth to notice that in your issue your go code actually is a slow request attack. Since that code opens a socket and does nothing. So it throttle your worker to the default slow request timeout. In real world you would want to deal with this attack with extra logic.

bobrik:
> Well the backpressure happens when holding a connection so either we force push it to worker or drop it.

Why accept a connection above configured limit in the first place?

> The attacker would be able to reject legit connections that are not actual slow requests since your worker thread are busy with their slow attack requests.

The problem is that attacker can open more connections than configured limit because limit is not fully enforced. The problem you're describing can happen either way.

fakeshadow:
> Why accept a connection above configured limit in the first place?

Because worker availability is atomicly checked and you don't want to iterator every worker's atomic state before accept a connection. For example if you have 32 workers you would need 32 atomic read of the worker state to decide if at certain point you can accept a connection or not. 

> The problem is that attacker can open more connections than configured limit because limit is not fully enforced. The problem you're describing can happen either way.

That's why I said you want extra logic to limit the remote peer from slow attack. You don't just drop connection like a blind man and say fuck it I don't care where it come from. 
The point here is you are using a slow request attack as an example of how legit accepting logic should work. I'm just pointing out they are different thing.

fakeshadow:
That said I'm having an idea of reduce the atomic cost but it would probably be another PR.

robjtede (approved):

fakeshadow:
There is issue with this change.
As backlogs are not drained on backpressure. Whoever bring the server back from backpressure should trigger accpet metond on every socket. Otherwise the backlog could be stucked until another new request comes in.

PR #323, opened 2021-04-09
Separate WorkerHandle to two parts

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Separate `WorkerHandle` into two parts as `WorkerHandleAccept` and `WorkerHandleServer`.

This would remove the need of `Clone` of worker handle and make it possible to adopt high performance spsc channel for server/accept to worker messaging. (Ring buffers are tend to be not (cheaply) `Clone`)

Remove `WorkerCommand` new type and use `Conn` directly.


robjtede (commented):
nice logical split

robjtede (approved):

PR #326, opened 2021-04-10
add spsc channel

### LINKED ISSUE #320: Use async spsc channel for dispatching workload.
Currently mpsc channel is used in actix-server to distribute connection to worker thread and there is extra overhead.

In case where accepting rate matters(Large amount of short lived connections) this could be a perf impacting factor. 

Therefore if there is a safe/fast unbounded spsc channel support cross thread task wake up through `std::task::Waker` it could be a worthy new dep to be introduced to actix-net.

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Use spsc channel for dispatch message from Accetp/ServerBuilder to workers. This would improve performance on accepting connections in general. But it's only impactful when managing short lived connections.(no keep-alive)

That said there is also a trade off to this change. One additional dep have been introduced and only used in actix-server.

Example bench code:
```rust
use std::io;

use actix_rt::net::TcpStream;
use actix_server::Server;
use actix_service::fn_service;
use tokio::io::AsyncWriteExt;

const BUF: &[u8] =
    b"HTTP/1.1 200 OK\r\n\
    content-length: 12\r\n\
    connection: close\r\n\
    date: Thu, 01 Jan 1970 12:34:56 UTC\r\n\
    \r\n\
    Hello World!\
    ";

fn main() -> io::Result<()> {
    let name = "short_life";
    let addr = "127.0.0.1:8080";

    actix_rt::System::new().block_on(async {
        Server::build()
            .bind(name, addr, || fn_service(response))?
            .run()
            .await
    })
}

async fn response(mut stream: TcpStream) -> io::Result<()> {
    stream.write(BUF).await?;
    stream.flush().await?;
    stream.shutdown().await
}
```

Benchmark tool and config:
`./wrk -t12 -c512 -d10 http://localhost:8080/`

Sample result(Result could be vary from run to run and differ on different system/metal):
```
refactor/spsc branch:

  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.44ms    3.85ms  55.50ms   88.34%
    Req/Sec    13.45k     3.56k   25.58k    73.11%
  1609624 requests in 10.09s, 164.25MB read
Requests/sec: 159509.95
Transfer/sec:     16.28MB


actix-server = "2.0.0-beta.4":

  12 threads and 512 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     2.82ms    4.71ms  69.68ms   89.25%
    Req/Sec    13.28k     3.03k   24.04k    73.65%
  1586311 requests in 10.10s, 161.87MB read
Requests/sec: 157076.48
Transfer/sec:     16.03MB


actix-server = "1.0.4":

  12 threads and 512 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     3.00ms    4.10ms  51.04ms   86.92%
    Req/Sec    12.79k     2.46k   24.48k    73.30%
  1534088 requests in 10.09s, 156.54MB read
Requests/sec: 152010.34
Transfer/sec:     15.51MB
```

Closes #320

### DIFF
--- a/actix-server/Cargo.toml
+++ b/actix-server/Cargo.toml
@@ -17,18 +17,17 @@ edition = "2018"
 name = "actix_server"
 path = "src/lib.rs"
 
-[features]
-default = []
-
 [dependencies]
 actix-rt = { version = "2.0.0", default-features = false }
 actix-service = "2.0.0-beta.5"
 actix-utils = "3.0.0-beta.4"
 
 futures-core = { version = "0.3.7", default-features = false, features = ["alloc"] }
+futures-util = { version = "0.3.7", default-features = false }
 log = "0.4"
-mio = { version = "0.7.6", features = ["os-poll", "net"] }
+mio = { version = "0.7.11", features = ["os-poll", "net", "os-util"] }
 num_cpus = "1.13"
+rtrb = ""
 slab = "0.4"
 tokio = { version = "1.2", features = ["sync"] }
 
--- a/actix-server/Cargo.toml
+++ b/actix-server/Cargo.toml
@@ -23,11 +23,10 @@ actix-service = "2.0.0-beta.5"
 actix-utils = "3.0.0-beta.4"
 
 futures-core = { version = "0.3.7", default-features = false, features = ["alloc"] }
-futures-util = { version = "0.3.7", default-features = false }
 log = "0.4"
-mio = { version = "0.7.11", features = ["os-poll", "net", "os-util"] }
+mio = { version = "0.7.11", features = ["os-poll", "net"] }
 num_cpus = "1.13"
-rtrb = ""
+rtrb = "0.1"
 slab = "0.4"
 tokio = { version = "1.2", features = ["sync"] }
 
--- a/actix-server/src/worker.rs
+++ b/actix-server/src/worker.rs
@@ -22,7 +22,7 @@ use tokio::sync::oneshot;
 
 use crate::service::{BoxedServerService, InternalServiceFactory};
 use crate::socket::MioStream;
-use crate::spsc::{channel, Receiver, Sender};
+use crate::spsc;
 use crate::waker_queue::{WakerInterest, WakerQueue};
 use crate::{join_all, Token};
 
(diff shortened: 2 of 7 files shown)


actix-server/Cargo.toml
```diff
@@ -17,18 +17,17 @@ edition = "2018"
...
-[features]
-default = []
-
 [dependencies]
 actix-rt = { version = "2.0.0", default-features = false }
 actix-service = "2.0.0-beta.5"
 actix-utils = "3.0.0-beta.4"
 
 futures-core = { version = "0.3.7", default-features = false, features = ["alloc"] }
+futures-util = { version = "0.3.7", default-features = false }
```
robjtede, line 26: if it's just for task mod could be futures-task

actix-server/Cargo.toml
```diff
@@ -17,18 +17,17 @@ edition = "2018"
...
 log = "0.4"
-mio = { version = "0.7.6", features = ["os-poll", "net"] }
+mio = { version = "0.7.11", features = ["os-poll", "net", "os-util"] }
 num_cpus = "1.13"
+rtrb = ""
```
robjtede, line 30: version

actix-server/src/worker.rs
```diff
@@ -18,19 +18,17 @@ use actix_rt::{
...
 use log::{error, info, trace};
-use tokio::sync::{
-    mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
-    oneshot,
-};
+use tokio::sync::oneshot;
 
 use crate::service::{BoxedServerService, InternalServiceFactory};
 use crate::socket::MioStream;
+use crate::spsc::{channel, Receiver, Sender};
```
robjtede, line 25: for readability can the spsc stuff have the spsc namespace at call sites

actix-server/src/worker.rs
```diff
@@ -195,12 +193,13 @@ impl ServerWorkerConfig {
...
         idx: usize,
+        backlog: u32,
         factories: Vec<Box<dyn InternalServiceFactory>>,
         availability: WorkerAvailability,
         config: ServerWorkerConfig,
     ) -> (WorkerHandleAccept, WorkerHandleServer) {
-        let (tx1, rx) = unbounded_channel();
-        let (tx2, rx2) = unbounded_channel();
+        let (tx1, rx) = channel(backlog as _);
+        let (tx2, rx2) = channel(1);
```
robjtede, line 202: these could use better names

actix-server/Cargo.toml
fakeshadow: Still not decided what crate to use. There is no obvious choice for now.

robjtede: it's all internal so wouldn't worry, the interface over the channel impl seems reasonable

actix-server/Cargo.toml
fakeshadow: futures-task does not have `AtomicWaker`. That said it's exported with futures-core just doc hidden. So only futures-core is enough here.

actix-server/Cargo.toml
fakeshadow: It's mostly about the bounded nature. The current server logic could possible overflow channel in edge case and bring down a heavy loaded worker thread which could be a serious issue.

PR #328, opened 2021-04-11
add test for restart worker thread

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Add test for restart worker thread.

### DIFF
--- a/actix-server/src/worker.rs
+++ b/actix-server/src/worker.rs
@@ -135,13 +135,6 @@ pub(crate) struct ServerWorker {
     shutdown_timeout: Duration,
 }
 
-/// Set worker to unavailable when dropping.
-impl Drop for ServerWorker {
-    fn drop(&mut self) {
-        self.rx.close();
-    }
-}
-
 struct WorkerService {
     factory: usize,
     status: WorkerServiceStatus,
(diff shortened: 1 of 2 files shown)


fakeshadow:
There is some additonal finding from this test.
Like a panic dropping worker thread have a race condition where it does not prevent new connection to be accepted. 

I believe the PR should be merged first and the problems can be looked into later.

actix-server/src/worker.rs
```diff
@@ -135,6 +135,13 @@ pub(crate) struct ServerWorker {
     shutdown_timeout: Duration,
 }
 
+/// Set worker to unavailable when dropping.
+impl Drop for ServerWorker {
+    fn drop(&mut self) {
+        self.rx.close();
```
robjtede, line 141: This already happens when the receiver is dropped, no?

fakeshadow: The drop order could be different than explicit call close and we want Accept to know the close asap. Like every cycle matters.

robjtede: the drop order is well defined in rust, rx is the first field and will be dropped first

fakeshadow: Future maintainer could be change the order and accidently make the time window larger.
And `UnboundedReciever` drop does not call close immediately. It does an atomic read first and return if there is no sender anymore. It's an optimization on the channel part but it make the close a bit later.

robjtede: yep, if you think it's important to make this explicit then we'll keep it

fakeshadow: I have added a note for ServerWorker field and removed the close in drop

robjtede (approved):

Issue #327, opened 2021-04-11
Use single atomic for tracking all workers availability.


fakeshadow:
In actix-server every worker has a separate availability tracker as atomic boolean. This makes tracking the availability of all workers hard.

A single `AtomicUsize` can be used to track all worker availability with every worker take one bit. This would reduce the atomic cost of tracking worker state and enable cheap hint before accepting connections. (When all workers are down accept would be stopped). 

This would also help to resolve bug like #329.

robjtede:
only supporting up to 64 workers would not be an acceptable limitation of using a single atomic

fakeshadow:
Yea. On a second though usize could not be a good choice as on 32bit system this would limit one group to 32.
So we can group every 64 workers to one atomic u64 and append addition worker groups to a vector.

That said the atomic cost would grow linear with the worker group count if accept would read all groups before accepting.

There is another approach to this is we eat some delay and potential over accepting and use `mio`'s waker to report worker un availbility state through the waker queue. This could eliminate atomic cost at happy path but would cause the worker state change delayed with at least one extra poll.

fakeshadow:
Close as #343 would resolve this issue.

PR #332, opened 2021-04-14
Fix bug where backpressure happen too early

### LINKED ISSUE #329: Accept go into backpressure early in edge case
This is a bug accidently caught by https://github.com/actix/actix-net/pull/328.

When server starts with n worker if worker n goes into faulty state on accepting connection it would trigger server backpressure even if all n-1 workers are in good state.

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

When accepting connections there was only one round trip for checking the availability state of workers from Accept.next to Accept.next + Accept.handles.len().
This would trigger backpressure early and force drain backlog to workers from the second round regardless their availability states.

This PR use [u128; 4] to track up to 512 worker handles' state and only enter backpressure when all of them goes into non available state.

Closes #329

### DIFF
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -510,3 +510,63 @@ impl Accept {
         self.avail.set_available(idx, false);
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::Availability;
+
+    fn single(aval: &mut Availability, idx: usize) {
+        aval.set_available(idx, true);
+        assert!(aval.available());
+
+        aval.set_available(idx, false);
+        assert!(!aval.available());
+    }
+
+    fn multi(aval: &mut Availability, mut idx: Vec<usize>) {
+        idx.iter().for_each(|idx| aval.set_available(*idx, true));
+
+        assert!(aval.available());
+
+        while let Some(idx) = idx.pop() {
+            assert!(aval.available());
+            aval.set_available(idx, false);
+        }
+
+        assert!(!aval.available());
+    }
+
+    #[test]
+    fn availability() {
+        let mut aval = Availability::default();
+
+        single(&mut aval, 1);
+        single(&mut aval, 128);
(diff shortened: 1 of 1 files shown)


fakeshadow:
a simpler double loop can do the same job. But with https://github.com/actix/actix-net/issues/327 in mind this could be a better approach to get rid of the atomic bool operation from all work threads.

actix-server/src/accept.rs
```diff
@@ -83,9 +83,59 @@ struct Accept {
     handles: Vec<WorkerHandleAccept>,
     srv: Server,
     next: usize,
+    avail: Availability,
     backpressure: bool,
 }
 
+/// Array of u128 with every bit as marker for a worker handle's availability.
+struct Availability([u128; 4]);
```
robjtede, line 91: any reason to hand roll this as opposed to using one of the structures from `bitvec` ?

actix-server/src/accept.rs
```diff
@@ -83,9 +83,59 @@ struct Accept {
...
+        let (offset, idx) = if idx < 128 {
+            (0, idx)
+        } else if idx < 128 * 2 {
+            (1, idx - 128)
+        } else if idx < 128 * 3 {
+            (2, idx - 128 * 2)
+        } else if idx < 128 * 4 {
+            (3, idx - 128 * 3)
+        } else {
+            unreachable!("Max WorkerHandle count is 512")
```
robjtede, line 116: would be a panic not an unreachable unless worker count was specified using `u8`s

actix-server/src/accept.rs
fakeshadow: We may want to replace this with Arc<[AtomicU64]> when #327 progress. The current one is mainly to fix the backpressure bug and find a suitable API for it.

robjtede (approved):
makes sense

PR #333, opened 2021-04-15
Fix worker are notified to stop with non_graceful shutdown

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

With non graceful shutdown there is no signal for notify worker threads to exit. This could result in leak of worker threads.

This PR changes to notify worker threads to shutdown on all cases. But only in graceful shutdown case server builer wait for worker shutdown response.

### DIFF
--- a/actix-server/src/builder.rs
+++ b/actix-server/src/builder.rs
@@ -381,45 +381,29 @@ impl ServerBuilder {
                 let notify = std::mem::take(&mut self.notify);
 
                 // stop workers
-                if !self.handles.is_empty() && graceful {
-                    let iter = self
-                        .handles
-                        .iter()
-                        .map(move |worker| worker.1.stop(graceful))
-                        .collect();
+                let stop = self
+                    .handles
+                    .iter()
+                    .map(move |worker| worker.1.stop(graceful))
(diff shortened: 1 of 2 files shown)


fakeshadow:
This changes the behavior of server as force shutdown was not exiting worker threads. There could be people using it as a feature rather than bug.

actix-server/src/builder.rs
```diff
@@ -381,45 +381,29 @@ impl ServerBuilder {
...
                         let _ = tx.send(());
                     }
                     for tx in notify {
                         let _ = tx.send(());
                     }
-                }
+
+                    if exit {
+                        sleep(Duration::from_millis(300)).await;
+                        System::current().stop();
```
robjtede, line 404: should we code this as try_current ?

fakeshadow: Right now we still expect system when running a server so I think it's fine here.
We can change when actix-rt is made optional.

fakeshadow: The thing I have no idea is why there is 300 ms delay.

robjtede (approved):

PR #335, opened 2021-04-15
remove pipeline from public api

robjtede:
## PR Type

Removal

## PR Checklist

## Overview
Pipeline doesn't seem to be used anywhere except in service tests. So remove it from public API so it may be removed completely later (or re-introduced if needed).


fakeshadow (approved):
It's useful for constructing service factories so there could be people using this somewhere.(Like barebone project for testing or something)

robjtede:
i don't think it will be a problem to just use fn_service / fn_factory now that and_then is exposed on Service{Factory}Ext

PR #336, opened 2021-04-16
Fix worker_avail

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Fix a bug introduced by #332

It's possible to trigger the assert condition in non debug build. Use an extra condition branch would prevent from double operation in all case.

### DIFF
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -116,12 +116,11 @@ impl Availability {
             panic!("Max WorkerHandle count is 512")
         };
 
+        let off = 1 << idx as u128;
         if avail {
-            self.0[offset] |= 1 << idx as u128;
+            self.0[offset] |= off;
         } else {
-            let shift = 1 << idx as u128;
-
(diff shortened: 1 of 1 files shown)


actix-server/src/accept.rs
```diff
@@ -121,9 +121,9 @@ impl Availability {
         } else {
             let shift = 1 << idx as u128;
 
-            debug_assert_ne!(self.0[offset] & shift, 0);
-
-            self.0[offset] ^= shift;
+            if self.0[offset] & shift != 0 {
+                self.0[offset] ^= shift;
+            }
```
robjtede, line 126: pure bitwise set to zero ?

```suggestion
            self.0[offset] & ~shift
```

robjtede: oh what rust syntax is ! for this ?

fakeshadow: yea. fixed.

PR #340, opened 2021-04-16
Fix memory ordering of WorkerAvailability

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Remove duplicate worker idx field from `WorkerHandleAccept` as `WorkerAvailability` already hold one and it can be used by handle.

Use stronger memory ordering for `WorkerAvailability::set`. In `<ServerWorker as Future>::poll` there is recursive call happens that would call set multiple times. A stronger ordering would make sure these calls happens in the order we want. (Worker availability change is a cold path so the cost of strong atomic should not be a problem and using SeqCst can also be considered)

Explicitly set `WorkerAvailability`'s atomic boolean to true when dropping `ServerWorker`. `Accept` would only notice an unexpected worker shutdown if it consider worker available and the removal/re-start of worker happens after a failed `send_connection` only.

### DIFF
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -128,7 +128,7 @@ impl Availability {
     /// This would result in a re-check on all workers' availability.
     fn set_available_all(&mut self, handles: &[WorkerHandleAccept]) {
         handles.iter().for_each(|handle| {
-            self.set_available(handle.idx, true);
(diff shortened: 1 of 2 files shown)


robjtede (approved):
> using SeqCst can also be considered

as I understand it, SeqCst is not necessary for the swap

PR #343, opened 2021-04-17
refactor connection counter

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Refactor connection counter to use `Arc<AtomicUsize>`. This counter is increment by `Accept` and decrement by `ServerWorker` to keep the counter in sync between threads for figuring out the connection limit of every worker.

Compared to current hint base limit management this would have additional atomic overhead in the form of dropping the `Arc<AtomicUsize>` when counter guard goes out of scope(Where the decrement of it happen). In return it gives a strong sync on the counter and make `Accept`'s backpressure timing more accurate.

Remove `slab` from dep.


fakeshadow:
Example Code:

```rust
use std::{
    io,
    net::SocketAddr,
    sync::atomic::{AtomicUsize, Ordering},
    time::Duration,
};

use actix_rt::net::TcpStream;
use actix_server::Server;
use actix_service::fn_service;
use log::info;
use tokio::io::{AsyncReadExt, AsyncWriteExt};

const BUF: &[u8] = b"HTTP/1.1 200 OK\r\n\r\n";

#[actix_rt::main]
async fn main() -> io::Result<()> {
    std::env::set_var("RUST_LOG", "actix=trace, info");
    env_logger::init();

    let name = "backpressure";
    let addr = "127.0.0.1:8080"
        .parse()
        .map_err(|e| io::Error::new(io::ErrorKind::AddrNotAvailable, e))?;

    let server = Server::build()
        .backlog(10)
        .maxconn(2)
        .workers(2)
        .bind(name, addr, || fn_service(response))?
        .run();

    let conns = client(addr).await?;
    let mut buf = [0u8; 1024];
    for mut conn in conns {
        let _ = conn.read(&mut buf).await.unwrap();
        conn.flush().await.unwrap();
        conn.shutdown().await.unwrap();
    }

    info!("All client requests are dropped");

    server.stop(true).await;

    Ok(())
}

static COUNTER: AtomicUsize = AtomicUsize::new(1);

async fn response(mut stream: TcpStream) -> io::Result<()> {
    info!(
        "Accepted total of {} connections",
        COUNTER.fetch_add(1, Ordering::Relaxed)
    );

    let mut buf = [0u8; 8];
    let _ = tokio::time::timeout(std::time::Duration::from_secs(3), stream.read(&mut buf)).await;
    stream.write(BUF).await?;
    stream.flush().await?;
    stream.shutdown().await
}

async fn client(addr: SocketAddr) -> io::Result<Vec<TcpStream>> {
    let mut conns = Vec::new();

    for _i in 0..30 {
        let conn = TcpStream::connect(addr).await?;
        conns.push(conn);
        actix_rt::time::sleep(Duration::from_millis(50)).await;
    }

    Ok(conns)
}
```

Expect log output:
```
[2021-04-19T16:36:36Z INFO  qa] Accepted total of 30 connections
[2021-04-19T16:36:39Z INFO  qa] All client requests are dropped
[2021-04-19T16:36:39Z INFO  actix_server::worker] Shutting down worker, 0 connections
[2021-04-19T16:36:39Z INFO  actix_server::accept] Paused accepting connections on 127.0.0.1:8080
[2021-04-19T16:36:39Z INFO  actix_server::worker] Shutting down worker, 0 connections
[2021-04-19T16:36:39Z INFO  actix_server::accept] Accept is stopped.

```

actix-server/src/worker.rs
```diff
@@ -38,35 +38,114 @@ pub(crate) struct Stop {
...
+        Self {
+            counter: Arc::new(AtomicUsize::new(1)),
+            limit,
+        }
+    }
+
+    /// Increment counter it by 1 and return false when hitting limit
+    #[inline(always)]
+    pub(crate) fn incr(&self) -> bool {
+        self.counter.fetch_add(1, Ordering::Relaxed) != self.limit
```
robjtede, line 78: condition should be on >= limit otherwise there's a race condition on the increment causing missed signalling of non-readiness.


would also suggest creating a simple loom test for it; my attempts seem to suggest that relaxed is okay here but it feels wrong

actix-server/src/worker.rs
```diff
@@ -38,35 +38,114 @@ pub(crate) struct Stop {
...
+    }
+
+    /// Decrement counter it by 1 and return true when hitting limit
+    #[inline(always)]
+    pub(crate) fn derc(&self) -> bool {
```
robjtede, line 83:
```suggestion
    pub(crate) fn dec(&self) -> bool {
```

actix-server/src/worker.rs
```diff
@@ -38,35 +38,114 @@ pub(crate) struct Stop {
...
+    pub(crate) fn new(limit: usize) -> Self {
+        Self {
+            counter: Arc::new(AtomicUsize::new(1)),
+            limit,
+        }
+    }
+
+    /// Increment counter it by 1 and return false when hitting limit
+    #[inline(always)]
+    pub(crate) fn incr(&self) -> bool {
```
robjtede, line 77:
```suggestion
    pub(crate) fn inc(&self) -> bool {
```

robjtede (changes requested):

actix-server/src/worker.rs
fakeshadow: There is no race as Accept is the only thread that increment it. The order does not really matter also because as soon as counter equals to limit a worker is marked as un avail.

We only have two thread operate on certain counter and there is no 3rd thread get in the way to inc or dec the counter.

actix-server/src/worker.rs
```diff
@@ -38,35 +38,114 @@ pub(crate) struct Stop {
...
+#[derive(Clone)]
+pub(crate) struct Counter {
+    counter: Arc<AtomicUsize>,
+    limit: usize,
+}
+
+impl Counter {
+    pub(crate) fn new(limit: usize) -> Self {
+        Self {
+            counter: Arc::new(AtomicUsize::new(1)),
```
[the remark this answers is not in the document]
fakeshadow: fetch_xxx atomic operation would wrap around on overflow. So unless we add a stronger order and do an assert there is no way to check overflow.

actix-server/src/worker.rs
```diff
@@ -38,35 +38,114 @@ pub(crate) struct Stop {
...
+        self.counter.load(Ordering::SeqCst) - 1
+    }
+}
+
+pub(crate) struct WorkerCounter {
+    idx: usize,
+    inner: Rc<(WakerQueue, Counter)>,
+}
+
+impl Clone for WorkerCounter {
```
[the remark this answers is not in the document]
fakeshadow: I use manual impl here to explicitly inform reader we want to Clone the RC and not what inside. Because WakerQueue and Counter are both Clone.

actix-server/src/worker.rs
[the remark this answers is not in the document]
robjtede: i didn't know about the wrapping behaviour

actix-server/src/worker.rs
robjtede: This makes sense. Would appreciate it being written up in the code so others can follow the logic later.

actix-server/src/worker.rs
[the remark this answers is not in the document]
fakeshadow: I can add an debug assert but I'm not sure if Relaxed order would work as we could have a lot of tasks on worker thread trying to decrement the counter.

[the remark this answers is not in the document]
robjtede: I think the syncronisation comes from elsewhere but I guess I'd feel more comfortable if they were AcqRel.

actix-server/src/worker.rs
fakeshadow: fixed

actix-server/src/worker.rs
fakeshadow: yea my english suck

actix-server/src/worker.rs
fakeshadow: Added some doc comment on the Counter type. Let me know if it's clear.

actix-server/src/worker.rs
```diff
@@ -38,35 +38,114 @@ pub(crate) struct Stop {
...
+        }
+    }
+
+    /// Increment counter it by 1 and return false when hitting limit
+    #[inline(always)]
+    pub(crate) fn incr(&self) -> bool {
+        self.counter.fetch_add(1, Ordering::Relaxed) != self.limit
+    }
+
+    /// Decrement counter it by 1 and return true when hitting limit
```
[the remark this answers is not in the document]
fakeshadow: fix

actix-server/src/worker.rs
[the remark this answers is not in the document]
fakeshadow: I would say we just use this and the worst case is over limiting cause a worker to hang.(I doubt it can happen but who knows). 

We can improve it when there is report on worker hang with very high(or little) workload coupled with very low worker limit.

robjtede (approved):

PR #344, opened 2021-04-17
Add System::spawn

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Add `System::spawn` to spawn async task on System runtime thread(s).

This is a short cut for `Arbiter::spawn(&self)` that would return a `JoinHandle` to be similar with most spawn functions.

Limitation:
No `actix_rt::spawn` is allowed inside.


fakeshadow:
I'm not sure about if this make sense. It's only useful for niche case where multi-threaded tokio is used as System runtime and with multiple indirection overhead.

PR #345, opened 2021-04-19
store Cow in actix-router Path

ibraheemdev:
## PR Type

Feature

## PR Checklist

## Overview
Store a cow in `Path`/`PathItem` to allow for owned strings as names/values.

actix/actix-web#1921


ibraheemdev:
Alternatives to `IntoIter` (which requires bumping the MSRV)

- array backed vec (which crate?)
- `Option::take`

actix-router/src/path.rs
```diff
@@ -95,18 +102,24 @@ impl<T: ResourcePath> Path<T> {
...
+                name.into(),
+                PathItem::Segment(self.skip + begin, self.skip + end),
+            )),
         }
     }
 
     #[doc(hidden)]
-    pub fn add_static(&mut self, name: &'static str, value: &'static str) {
-        self.segments.push((name, PathItem::Static(value)));
+    pub fn add_static(
```
ibraheemdev, line 116: I'm not sure what this should be called now...

neoeinstein: I think `add_static` is probably still fine. It effectively moves from meaning the `'static` lifetime to the `'static` type constraint.

actix-router/src/resource.rs
```diff
@@ -274,15 +274,15 @@ impl ResourceDef {
             PatternType::Dynamic(ref re, ref names, len) => {
                 let mut idx = 0;
                 let mut pos = 0;
-                let mut segments: [PathItem; MAX_DYNAMIC_SEGMENTS] =
-                    [PathItem::Static(""); MAX_DYNAMIC_SEGMENTS];
+                let mut segments: [Option<PathItem>; MAX_DYNAMIC_SEGMENTS] = Default::default();
```
thalesfragoso, line 277: Maybe we can get rid of the `Option` and the `Default` implementation by using something like:
```rust
const PATH_ITEM = PathItem::Static(Cow::Borrowed(""));
let mut segments = [PATH_ITEM; MAX_DYNAMIC_SEGMENTS];
```

thalesfragoso (commented):
Thanks for the PR, I left some "weird" trick to maybe improve it.

actix-router/src/resource.rs
ibraheemdev: Arrays (in safe code) have to initialize all their elements, what's wrong with the `Default` implementation? Also, I guess the `Option` could be replaced with `mem::take` because `PathItem` implements `Default`. Ideally we could use `IntoIter`, but msrv is still 1.46.

thalesfragoso: That's a trick to be able to use a non-Copy type to initialize arrays like that, all elements will be initialized.

ibraheemdev: You don't initialize anything there.

ibraheemdev: Yes, I know, but how is that better than `Default`?

thalesfragoso: See:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=01f097a652dd712fd85126309e0c9abb

ibraheemdev: I understand it works :), but why is it better than `Default`?

ibraheemdev: To get rid of the options we need `Default` anyways.

thalesfragoso: I guess I don't understand why you added an `Option` there.

ibraheemdev: We can't use `IntoIter` because of the minimum supported rust version. The other (safe) way to iterate over an array by value is to iterate over mutable references, and then replace it with a default value. `Option::take` is a nice way to do this. We could use `mem::take` instead, which requires `Default`.

thalesfragoso: Yes, sorry, I think I kinda got stuck in the array initialization and forgot about the iteration, `mem::take` seems like a better option (pun intended) though IMO.

actix-router/src/resource.rs
```diff
@@ -294,8 +294,12 @@ impl ResourceDef {
...
                 }
-                for idx in 0..idx {
-                    path.add(names[idx], segments[idx]);
+                for (idx, segment) in segments
+                    .iter_mut()
+                    .enumerate()
+                    .take_while(|(i, _)| *i != idx)
+                {
+                    path.add(names[idx], segment.take().unwrap());
                 }
```
aliemjay, line 303: For functional style:
```suggestion
                segments
                    .iter_mut()
                    .flat_map(Option::take)
                    .zip(names)
                    .for_each(|(segment, name)| path.add(name, segment));
```
 for  procedural:
```suggestion
                for idx in 0..names.len() {
                    path.add(names[idx], segments[idx].take().unwrap());
                }
```

I find the procedural more readable here.
In both cases. `idx` is not used!

aliemjay (commented):
Looks good to me overall.

Some notes on style...

actix-router/src/resource.rs
ibraheemdev: Good point, it short circuits on failure, so we can just use all the names here.

actix-router/src/path.rs
ibraheemdev: That makes sense, but I still don't really like it.

actix-router/src/path.rs
```diff
@@ -95,18 +102,24 @@ impl<T: ResourcePath> Path<T> {
...
+        &mut self,
+        name: impl Into<Cow<'static, str>>,
+        value: impl Into<Cow<'static, str>>,
+    ) {
+        let value: Cow<'static, str> = value.into();
```
robjtede, line 121: isn't this annotation unnecessary?

robjtede (approved):

robjtede:
just needs that changelog entry

robjtede:
released in actix-router 0.4.0

PR #348, opened 2021-04-23
Refactor/actix server waker

fakeshadow:
## PR Type

Refactor

## PR Checklist

## Overview

Rework actix-server to use `std::task::wake` trait for mio's poll thread wakeup.
This would enable calling poll based future types in mio's thread and pass mio's waker through Context while at the same time keep the code in safe without needing to mess with raw wakers and vtables.


fakeshadow:
This would need MSRV rise to 1.51

PR #350, opened 2021-04-26
Fix Display impl of MioListener

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Fix Display of `MioListener`. It would display both addr and fd.

Remove `SocketAddr` field from `ServerSocketInfo`. Reduce the size by more than 100 bytes.


JohnTitor (approved):
LGTM.

PR #355, opened 2021-04-30
fix race condition in actix-server graceful shutdown

fakeshadow:
## PR Type

Bug Fix

## PR Checklist

## Overview

Expose the race condtion by modify test.

### DIFF
--- a/actix-server/src/worker.rs
+++ b/actix-server/src/worker.rs
@@ -192,10 +192,112 @@ impl WorkerHandleServer {
     }
 }
 
-/// Service worker.
+/// Server worker.
 ///
 /// Worker accepts Socket objects via unbounded channel and starts stream processing.
 pub(crate) struct ServerWorker {
+    worker: Worker,
+    state: WorkerState,
+}
+
+impl ServerWorker {
+    fn new(
+        rx: UnboundedReceiver<Conn>,
+        rx2: UnboundedReceiver<Stop>,
+        counter: WorkerCounter,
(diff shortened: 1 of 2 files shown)


fakeshadow:
Added fix for the race condition. The fix would change worker handling logic.

Previously `ServerWorker` would check `UnboundedReceiver<Stop>` everytime `WorkerState` is mutated. This would have an early timing of going into `WorkerState::Shutdown` variant. 

With the fix `UnboundedReceiver<Stop>` would be checked only when `ServerWorker` itself is polled as a Future. This would delay the timing of `WorkerState::Shutdown` making it always happen after `UnboundedReceiver<Conn>` returns Pending state.

In general this can delay the shutdown duration by a bit when server is busy but at the same time it's a correct way of handling the shutdown and avoid race condition.

The reason behind this is when server is shutting down. `Accept` thread and `ServerWorker` thread(s) are signal to shutdown in a non-blocking manner. This means on `ServerWorker` shutdown `Accept` could still be running and accept new connections which is pushed to `ServerWorker`.

Reference:
https://github.com/actix/actix-net/blob/983abec77d3d57e13aaa4773e23befd1643bf914/actix-server/src/builder.rs#L350
https://github.com/actix/actix-net/blob/983abec77d3d57e13aaa4773e23befd1643bf914/actix-server/src/builder.rs#L354

fakeshadow:
Added back 500ms delay for `test_listen`. This is another race condition of server start up which would be fixed by #266

PR #357, opened 2021-06-01
router: don't decode %25 to '%'

aliemjay:
## PR Type

Bug fix

## PR Checklist

## Overview

Although the [documentation](https://actix.rs/docs/url-dispatch/) says that `path will be URL-unquoted and decoded into valid unicode string before matching pattern` , this is not the case - the matched url path segment may still have some percent encoded characters (specifically, `+` and `/`) .

Withh that given, `%25` should not unquoted to `%` to guarantee that matched segment is always a valid percent-encoded string so that the user can decode it himself. Consider the case in the linked issue!

Closes actix/actix-web#1157

### DIFF
--- a/actix-router/CHANGES.md
+++ b/actix-router/CHANGES.md
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
+* When matching URL parameters, `%25` is kept in the percent-encoded form - no longer decoded to `%`. [#357]
+
+[#357]: https://github.com/actix/actix-net/pull/357
 
 
 ## 0.2.7 - 2021-02-06
--- a/actix-router/src/url.rs
+++ b/actix-router/src/url.rs
@@ -204,24 +204,59 @@ mod tests {
     use super::*;
     use crate::{Path, ResourceDef};
 
+    const PROTECTED: &[u8] = b"%/+";
+
+    fn match_url(pattern: &'static str, url: impl AsRef<str>) -> Path<Url> {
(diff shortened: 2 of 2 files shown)


JohnTitor (approved):
Looks reasonable to me 👍

robjtede:
Havent thought about the implications yet but this would definitely need tests (and changelog).

aliemjay:
The only implications I can think of are: url parameters the user gets from the `Path` extractor will now have `%` encoded as `%25`. Custom regex users (i.e. `{xxx:.*}`) are also affected 

This is a breaking change but, I think, is necessary for correctness.

JohnTitor:
Since this changes the behavior anyway, it's definitely worth mentioning it on the changelog, even for correctness. I'm not sure about tests though, if you have some thoughts to harden it, it'd be great.

aliemjay:
Added more tests and changelog.

Should this be mentioned in the migration guide?

aliemjay:
Giving it more thought...

The question is does this really fixes the issue? Do we really want the user deal with percent-encoding in url parameters? I think thist should be totally abstracted away from the user!

JohnTitor:
> The question is does this really fixes the issue? Do we really want the user deal with percent-encoding in url parameters? I think thist should be totally abstracted away from the user!

It definitely needs a decision (actix/actix-web#2248). I think resolving mixed behavior is a straightforward way for now.

> Should this be mentioned in the migration guide?

I think it's okay not to mention.

JohnTitor:
Thanks!

PR #360, opened 2021-06-03
router: handle newline char '\n' in url

aliemjay:
## PR Type

Bug fix

## PR Checklist

## Overview

A `Path` instance may contain newline chars (`\n`) in the path segments. In such case, tests have shown some problems when matching URL parameters (specifically, path tail patterns; see the test case!)  . This turned out to be because in `regex`, `.` doesn't match `\n` by default.

The solution is to either keep newline chars in percent encoding (`%0A`) or to use regex flags to get the desired behavior. I tried to implement the latter.

### DIFF
--- a/actix-router/src/resource.rs
+++ b/actix-router/src/resource.rs
@@ -817,6 +817,32 @@ mod tests {
         assert!(re.is_match("/user/2345/sdg"));
     }
 
+    #[test]
+    fn test_newline() {
+        let re = ResourceDef::new("/user/a\nb");
+        assert!(re.is_match("/user/a\nb"));
+        assert!(!re.is_match("/user/a\nb/profile"));
+
+        let re = ResourceDef::new("/a{x}b/test/a{y}b");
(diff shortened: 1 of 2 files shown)


actix-router/src/resource.rs
```diff
@@ -10,6 +10,8 @@ use crate::{IntoPattern, Resource, ResourcePath};
 
 const MAX_DYNAMIC_SEGMENTS: usize = 16;
 
+const REGEX_FLAGS: &str = "(?s-m)";
```
robjtede, line 13: a doc comment for the readers who are not regex experts would be great

robjtede: or just reference the flags section in the regex crate somehow

robjtede (commented):
looks good, just a changelog entry and we can get this released and in testing along with next actix web beta

robjtede (approved):

robjtede:
released in actix-router 0.4.0

PR #366, opened 2021-06-27
router: fix multi-pattern and path tail matches

aliemjay:
## PR Type

Bug fix

## PR Checklist

## Overview

In muti-pattern, static patterns are accidentaly interpreted as regex. This is fixed by adding a flag to `parse()` (`force_dynamic`) to force it to always escape and render the pattern as regex even when it is static.

This also fixes the case when _static_ path tail (`/*`) is used after a dynamic segment:

>  `/static/*` ... already works
> `/{user}*` ... already works; stores the path tail in `user`
> `/{user}/*` ... doesn't work as expected.

Please see failing tests and linked issue.

Closes actix/actix-web#1647

### DIFF
--- a/actix-router/CHANGES.md
+++ b/actix-router/CHANGES.md
@@ -2,8 +2,11 @@
 
 ## Unreleased - 2021-xx-xx
 * Fix segment interpolation leaving `Path` in unintended state after matching. [#368]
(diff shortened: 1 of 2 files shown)


aliemjay:
@robjtede  I added one more commit of related refactoring. I'm unsure if you want it in another PR.

robjtede (approved):
great work

PR #365, opened 2021-06-26
`ResourceDef`: cleanup

aliemjay:
## PR Type

Refactor

## PR Checklist
Check your PR fulfills the following:

- [ ] Tests for the changes have been added / updated.
- [ ] Documentation comments have been added / updated.
- [ ] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt

## Overview

Simplify and remove duplicate code.

PR #368, opened 2021-06-27
`path.len()` != `path.path().len()`

aliemjay:
## PR Type

Bug Fix

## PR Checklist

## Overview

`Path::len()` has a very confusing name! 

This bug leaves the `Path` in  unintended  state after matching. It only applies to static non-prefix matches, which is always the last pattern to match. So, it doesn't have much of impact (except for users of `HttpRequest::match_info()` for actix-web). However, it caused more problems for me in #365 .

### DIFF
--- a/actix-router/CHANGES.md
+++ b/actix-router/CHANGES.md
@@ -1,6 +1,9 @@
 # Changes
 
 ## Unreleased - 2021-xx-xx
+* Fix segment interpolation leaving `Path` in unintended state after matching. [#368]
+
+[#368]: https://github.com/actix/actix-net/pull/368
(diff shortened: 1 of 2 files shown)


robjtede:
> Path::len() has a very confusing name!

Feel free to rename it to something more accurate and add docs if you think it would be helpful.

aliemjay:
> Feel free to rename it to something more accurate and add docs if you think it would be helpful.

I'll leave it for another PR. `Path` interface need more cleanup: `unprocessed() == path()` for example.

robjtede (approved):

PR #369, opened 2021-06-30
`spawn` should allow futures with non-unit outputs

ibraheemdev:
## PR Type
Feature

## PR Checklist


robjtede:
This is non-breaking, right?

ibraheemdev:
It's considered a minor breaking change.

robjtede:
Can you show me an example of a case where this would break code?

ibraheemdev:
I don't have an example, but I think it could potentially affect type inference. 

> The type of an argument to a function, or its return value, can be generalized to use generics, including by introducing a new type parameter (as long as it can be instantiated to the original type). ... Introducing generics in this way can potentially create type inference failures, but these are considered acceptable per the principles of the RFC: they only require local annotations that could have been inserted in advance.

https://rust-lang.github.io/rfcs/1105-api-evolution.html#minor-change-generalizing-to-generics-1

robjtede:
I suspect we're safe from inference failures because we're not actually introducing a new generic, we're just using the associated type of an existing generic.

ibraheemdev:
Yeah, that's probably true, but there could always be some quirky edge case that we didn't account for. `actix_rt::spawn` is used rarely for anything more than `spawn(foo)` or `let handle = spawn(foo); ... handle.await`, so I either way the risk of breakage is very small.

PR #373, opened 2021-07-17
rework resourcedef

robjtede:
## PR Type

Refactor + Docs + Improvements + Fixes

## PR Checklist

## Overview
- Codified spec for `ResourceDef`.
- Documented `ResourceDef`.
- Removed previously-deprecated methods on `ResourceDef`.
- Introduce `ResourceDef::pattern_iter` to get an iterator over all patterns in a multi-pattern resource.
- `ResourceDef::resource_path_from_iter` now takes an `IntoIterator`.
- Rename `ResourceDef::{is_prefix_match => find_match}`.
- Rename `ResourceDef::{match_path => capture_match_info}`.
- Rename `ResourceDef::{match_path_checked => capture_match_info_fn}`.
- Remove `ResourceDef::name_mut` and introduce `ResourceDef::set_name`.
- Return type of `ResourceDef::name` is now `Option<&str>`.
- Return type of `ResourceDef::pattern` is now `Option<&str>`.
- Rename `Router::{*_checked => *_fn}`.


aliemjay:
Great work. This is much better now.

aliemjay:
Since you are touching this, it would be nice to have documentation for empty patterns with something like

> `ResourceDef` supports empty patterns as both prefix and non-prefix patterns. This only matches an empty path.

I also found a bug with empty patterns `assert_eq!(ResourceDef::new("").find_match(""), Some(0))`

I think also it is necessay to have `is_prefix(&self) -> bool` to get complete matching info.

What do you think? I'll implement these myself if you like.

actix-router/src/resource.rs
```diff
@@ -112,82 +137,264 @@ impl ResourceDef {
...
+    /// // TODO: dyn set resource
+    /// // TODO: tail segment resource
+    /// ```
     #[inline]
     pub fn is_match(&self, path: &str) -> bool {
         profile_method!(is_match);
 
+        // in effect this function could be expressed as:
+        // self.find_match(path).is_some()
+
```
aliemjay, line 397: I think we have this because `Regex::is_match` is faster the `Regex::matches`

actix-router/src/resource.rs
```diff
@@ -112,82 +137,264 @@ impl ResourceDef {
...
         match self.pat_type {
             PatternType::Static(ref s) => s == path,
             PatternType::Prefix(ref s) => path.starts_with(s),
```
aliemjay, line 400:
```suggestion
            PatternType::Prefix(_) => self.find_match(s).is_some(),
```
This should have the logic to detect end of segment same as in `find_match`
Currently `ResourceDef::prefix("/user").is_match("/user123/abc")`!

actix-router/src/resource.rs
```diff
@@ -196,9 +403,37 @@ impl ResourceDef {
...
+    /// assert_eq!(resource.find_match("/user/123/"), Some(9));
+    /// assert_eq!(resource.find_match("/user/123/stars"), Some(9));
+    /// assert!(resource.find_match("/user/").is_none());
+    /// assert!(resource.find_match("/foo").is_none());
+    /// ```
+    pub fn find_match(&self, path: &str) -> Option<usize> {
+        profile_method!(find_match);
 
         let path_len = path.len();
         let path = if path.is_empty() { "/" } else { path };
```
aliemjay, line 439:
```suggestion
```
This should be removed to have a consistent behavior with `is_match` and `match_path`. This also fixes the bug with empty patterns I mentioned earlier.

actix-router/src/resource.rs
```diff
@@ -217,19 +452,19 @@ impl ResourceDef {
...
                         //
-                        // eg: Prefix("/test/") or Prefix("/test") would match:
-                        // - /test/foo
-                        // - /test/foo
+                        // eg: Prefix("/test/") or Prefix("/test") would match "/test/" and
+                        // "/test/foo" but Prefix("/test") would not process "/test" here since it
+                        // is handled by the earlier if case
 
                         if prefix.ends_with('/') {
                             prefix.len() - 1
```
aliemjay, line 470:
```suggestion
                        if prefix.ends_with('/') {
                            prefix.len()
```
I think this branch should also consume the full length of prefix. Otherwise we have inconsistent behavior with dynamic prefix `ResourceDef::prefix("/abc/").find_match("/abc/def") == Some(4)` vs `ResourceDef::prefix("/{id}/").find_match("/abc/def") == Some(5)`. I also think it is more intuitive behavior.

In this case this match arm would bre reduced to 
```rust
let rem = path.strip_prefix(prefix)?;

if prefix.ends_with('/') ||  rem.starts_with('/') || rem.is_empty() {
    Some(prefix.len())
} else {
    None
}
```

actix-router/src/resource.rs
```diff
@@ -254,38 +489,47 @@ impl ResourceDef {
...
-        profile_method!(match_path_checked);
+        profile_method!(is_path_match_fn);
 
-        let mut segments: [PathItem; MAX_DYNAMIC_SEGMENTS] = Default::default();
+        let mut segments = <[PathItem; MAX_DYNAMIC_SEGMENTS]>::default();
         let path = res.resource_path();
 
-        let (matched_len, matched_vars) = match self.pat_type {
+        let (matched_len, matched_vars, tail) = match self.pat_type {
             PatternType::Static(ref segment) => {
```
aliemjay, line 525:
```suggestion
        let (matched_len, matched_vars, tail) = match self.pat_type {
            PatternType::Static(_) | PatternType::Prefix(_) => match self.find_match(path.path()) {
                Some(len) => (len, None, None),
                None => return false, 
            }
```

at the cost of one call.

aliemjay (changes requested):
I'll take this chance to discuss some problems.

actix-router/src/resource.rs
robjtede: yep, i have a similar comment un-pushed

i've found discrepencies in the output of these functions around prefixes i'm reconciling too

actix-router/src/resource.rs
robjtede: should have read them all, this is the discrepency I found also

in your mind, should we allow prefixes to do this or should they be used only for `/`-delimited "scopes"?

aliemjay: I think this shouldn't be allowed. They should match full segments.

robjtede: okay, i was almost leaning the same way, i can't really think of a use case

aliemjay: I've just found that full segment match is not required for dynamic prefixes in `find_match` and `match_path_checked` either. So in web the default service for `web::scope("/{domain}/user")` would be called on url `/xxx/user123`. This seems wrong to me.

robjtede: yeah i agree that's weird, good to start defining the router properly

actix-router/src/resource.rs
```diff
@@ -922,15 +876,29 @@ pub(crate) fn insert_slash(path: &str) -> Cow<'_, str> {
     }
 }
 
+/// Returns true if `prefix` acts as a proper prefix (i.e., separated by a slash) in `path`.
+///
+/// The `strict` refers to the fact that this will return `false` if `prefix == path`.
+fn is_strict_prefix(prefix: &str, path: &str) -> bool {
```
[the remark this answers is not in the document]
robjtede: yep confirmed that this code works (and shouldnt):

```rust
```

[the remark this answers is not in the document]
robjtede: i'll finish this PR off and we'll come back to this, it's pretty overloaded anyway

actix-router/src/resource.rs
```diff
@@ -112,212 +289,446 @@ impl ResourceDef {
...
+    /// assert!(try_match(&resource, &mut path));
+    /// assert_eq!(path.get("id").unwrap(), "james");
+    /// assert_eq!(path.unprocessed(), "/stars");
+    ///
+    /// // path matches but fails check function; no segments are collected
+    /// let mut path = Path::new("/user/admin/stars");
+    /// assert!(!try_match(&resource, &mut path));
+    /// assert_eq!(path.unprocessed(), "/user/admin/stars");
+    /// ```
+    pub fn capture_match_info_fn<R, T, F, U>(
```
aliemjay, line 704: I think these unnecessary bounds should be relaxed and `user_data` removed
```rust
pub fn capture_match_info_fn<R, T, F>(&self, res: &mut R, check: F)
where
    ...
    F: FnOnce(&R) -> bool
```

aliemjay (approved):
Very nice. Much better now. This is one area where docs were severely lacking.

PR #374, opened 2021-07-17
add support for io-uring

fakeshadow:
## PR Type

Feature

## PR Checklist

## Overview

Initial io-uring support with `tokio-uring` crate dependency.
Only `actix_rt::Arbiter` is started with `tokio-uring` runtime. `actix_rt::System` is left out for now.

### DIFF
--- a/actix-router/src/path.rs
+++ b/actix-router/src/path.rs
@@ -122,7 +122,7 @@ impl<T: ResourcePath> Path<T> {
     /// Get matched parameter by name without type conversion
     pub fn get(&self, key: &str) -> Option<&str> {
         profile_method!(get);
-        
+
         for item in self.segments.iter() {
             if key == item.0 {
                 return match item.1 {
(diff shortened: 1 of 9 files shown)


fakeshadow:
This could be the first platform specific introduced. CI may need a rework and remove --all-features flag on cargo check/test

edward-shen:
Docs should be updated to reflect the availability of `io-uring` as well, e.g. in the first paragraph in `lib.rs`.

This should be a completely transparent change to downstream users, right? As in, `acitx-web` shouldn't need a major version bump to start using `io-uring` or opt into it for Linux targets?

fakeshadow:
> This should be a completely transparent change to downstream users, right? As in, `acitx-web` shouldn't need a major version bump to start using `io-uring` or opt into it for Linux targets?

io-uring feature is additive and the default feature would be disabling it.

edward-shen:
What's the status of this PR? @robjtede, did you determine if the CI failures were transient or a legitimate issue? The last commit to this branch seems to imply it was more of a testing issue than a race condition.

robjtede (approved):

PR #382, opened 2021-07-20
ResourceDef: always end match in a segment boundary

### LINKED ISSUE #376: dynamic prefixes should not be able match without a slash seperator
> Good! Now we have to find a solution for dynamic prefixes. One idea is to implement it in regex, where each dynamic prefix pattern is put inside a capture group with a pattern to detect segment boundary: `(original_regex_pattern)[/$]`. But this may add complexity. I'm not sure.

_Originally posted by @aliemjay in https://github.com/actix/actix-net/pull/373#discussion_r672394311_

test case that should not work but does:
```rust
```

aliemjay:
## PR Type

Bug Fix

## PR Checklist

## Overview

Co-authored by: @robjtede

An alternative implementation to #378 , which leaves `is_match` using the fast `Regex::is_match`. Although it may affect `find_match` performance by using `Regex::captures`.

Closes #376

### DIFF
--- a/actix-router/src/resource.rs
+++ b/actix-router/src/resource.rs
@@ -628,12 +628,12 @@ impl ResourceDef {
             PatternType::Prefix(prefix) if is_strict_prefix(prefix, path) => Some(prefix.len()),
             PatternType::Prefix(_) => None,
 
-            PatternType::Dynamic(re, _) => re.find(path).map(|m| m.end()),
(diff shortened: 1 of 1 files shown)


aliemjay:
I thought we could have a nicer solution, but unfortunately found out that `regex` crate doesn't support perl-style lookahead (otherwise we could have a fix with no change to the matching functions). Now I unsure if this is preferable to #378

aliemjay:
preceded by #383

PR #383, opened 2021-07-21
`ResourceDef` redefine behavior rigorously

### LINKED ISSUE #376: dynamic prefixes should not be able match without a slash seperator

aliemjay:
## PR Type

INSERT_PR_TYPE

## PR Checklist

## Overview

Please see the docs.

* panic if pattern doesn't start with slash
* prefix with a trailing slash now requires an empty segment
* ~fixed a bug where an empty prefix matches only empty path~
* Closes #376 

This shouldn't affect the behavior in actix-web because we always insert slash.

Tests were taken from #378 .

### DIFF
--- a/actix-router/src/resource.rs
+++ b/actix-router/src/resource.rs
@@ -28,9 +28,27 @@ const REGEX_FLAGS: &str = "(?s-m)";
 /// regex engine.
 ///
 ///
+/// # Pattern Format and Matching Behavior
+///
+/// Resource pattern is defined as a string of zero or more _segments_ where each segment is
+/// preceeded by a slash `/`.
+///
+/// This means that pattern string *must* either be empty or begin with a slash (`/`).
+/// This also implies that a trailing slash in pattern defines an empty segment.
+/// For example, the pattern `"/user/"` have two segment: `["user", ""]`
+///
+/// A key point to undertand is that `ResourceDef` matches segments, not string.
+/// It matches segments individually.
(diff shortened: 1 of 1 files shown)


robjtede:
> panic if pattern doesn't start with slash

this breaks external_resources in actix web

robjtede:
There's parts of this that I like but i'm going to go with #378 to solve #376 for now. It'll be a beta release so we've got time to revisit these points later.

aliemjay:
> > panic if pattern doesn't start with slash
> 
> this breaks external_resources in actix web

Then just restrict this definition to matching while still allowing such constructing such pattern?

I still believe this is a more correct, yet simpler behavior to reason about.  If we are to keep the old behavior we can have the magic in actix-web.

robjtede:
what i want to do actually, is clear up the router PRs here that are easy, release a beta, then move this crate to -web so we can catch breaks in web easier

robjtede:
> I still believe this is a more correct

I believe you. We'll get back to it :)

aliemjay:
superseded by actix/actix-web#2355

PR #388, opened 2021-09-25
Fix Service<u8> request documentation

asonix:
## PR Type
Documentation change

## Overview
Service documentation in actix-service/src/lib.rs still referred to Self::Request


robjtede (approved):
good spot, thanks

PR #396, opened 2021-10-18
Use tokio-rustls 0.23

edward-shen:
## PR Type

Feature

## PR Checklist

## Overview

This is related to actix/actix-web#2304, which requires `acitx-tls` to support `rustls` 0.20. `tokio-tls` 0.23 adds this support, so we just need to version bump everything, and address locations that are broken.

Note that a breaking change exists, as we no longer re-export `Session`.

Tests were made with `cargo t --all-features`, and verification of the change was performed with `cargo tree -p actix-tls --all-features | grep rustls`:
```
├── tokio-rustls v0.23.0
│   ├── rustls v0.20.0
├── rustls-pemfile v0.2.1
```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -22,4 +22,4 @@ ci-test-rt-linux = "    hack --feature-powerset test --package=actix-rt     --li
 ci-test-server-linux = "hack --feature-powerset test --package=actix-server --lib --tests --no-fail-fast -- --nocapture"
 
 # test lower msrv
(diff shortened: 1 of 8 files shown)


edward-shen:
Re: CI failures: rustls uses bool::then which were stabilized in 1.50 (rust-lang/rust#79299). Not sure if we can bump actix-net's MSRV to 1.50? (rustls real MSRV is 1.52.1 so optimally we'd bump to that instead).

robjtede:
1.52 is higher than i want to go :/ the higher level parts of actix have moved to 1.51 recently

PR #403, opened 2021-10-23
server bind accepts types impling regular service traits

robjtede:
## PR Type

Feature

## PR Checklist

## Overview
As title. This should help with fallible services.


actix-service/src/and_then.rs
```diff
@@ -261,6 +261,91 @@ where
     }
 }
 
+/// `.and_then_send()` service factory combinator
+pub struct AndThenSendServiceFactory<A, B, Req>
```
aliemjay, line 265: why not just use Arc in AndThenServiceFactory?

actix-service/src/and_then.rs
```diff
@@ -261,6 +261,91 @@ where
...
+
+    fn new_service(&self, cfg: A::Config) -> Self::Future {
+        AndThenServiceFactoryResponse::new(
+            self.inner_a.new_service(cfg.clone()),
+            self.inner_b.new_service(cfg),
+        )
+    }
+}
+
+impl<A: Clone, B: Clone, Req> Clone for AndThenSendServiceFactory<A, B, Req>
```
aliemjay, line 329: I'd rather put Clone bound in where clause. I had a hard time reading it :)

actix-server/src/builder.rs
```diff
@@ -147,91 +158,59 @@ impl ServerBuilder {
...
+    /// interfaces at the same time by passing a list of socket addresses.
+    pub fn bind<F, U, InitErr>(
+        mut self,
+        name: impl AsRef<str>,
+        addr: U,
+        factory: F,
+    ) -> io::Result<Self>
     where
-        F: ServiceFactory<TcpStream>,
+        F: ServiceFactory<TcpStream, Config = (), InitError = InitErr> + Send + Clone + 'static,
```
aliemjay, line 173: I thought of IntoServiceFactory to support passing closures directly, but I think using fn_service and fn_factory explicitly is better.

aliemjay (changes requested):
Nice to ditch ServiceFactoryFactory!

actix-service/src/and_then.rs
robjtede: It just seems unfortunate to have to downgrade the performance of and_then. I'm still tinkering though.

aliemjay: Since ServiceFactory::new_service is O(1) during app lifetime, I think it is reasonable to have more cost there to guarantee `Send`abliity. The output service type can keep using Rc shich is what matters!

aliemjay: of course, this would be an extra pointer indirection per-request Rc<Arc<>> :(

robjtede:
Trying to integrate this with Actix Web is proving to be the blocker. De-prioritizing this until next major version of -server to speed up actix-web v4 release.

PR #409, opened 2021-11-04
`Framed::poll_ready` flushes when buffer is full

### LINKED ISSUE #406: Codec: Framed::poll_ready does not register waker
It returns `Poll::Pending` without registering the waker:
https://github.com/actix/actix-net/blob/1c8fcaebbc2f385acece09f7fcc2fd565daae1c4/actix-codec/src/framed.rs#L303-L309

This might be the cause of https://github.com/actix/actix/issues/431, as noted by fakeshadow.

CC @elpiel

brockelmore:
## PR Type
Bug fix

## PR Checklist

## Overview
Before, a `SinkWriteFuture` wouldn't be woken up during a `poll_ready` call, if the buffer length was greater than the `HW` (high water mark). This fixes that by flushing the buffer if that is the case. It adds tests based on `tokio`'s `FramedWrite`.

This is a non-breaking change.

Fixes actix/actix#431, fixes #406


robjtede:
actix-codec 0.4.1 is released with this fixed, should just be a cargo update to get it

elpiel:
Awesome! Thanks for the fix @brockelmore !

PR #418, opened 2021-11-14
Circleci project setup

robjtede:
## PR Type

CI

## Overview
Want to trial CircleCI by running the io-uring tests on it.


robjtede:
nvm circleci machine runners do not have recent enough kernel for this

PR #424, opened 2021-12-02
Server: hide internal structure

aliemjay:
## PR Type

Refactor

## PR Checklist
Check your PR fulfills the following:

- [x] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt

## Overview
Hide implementation details of `Server`.

PR #426, opened 2021-12-04
Server: run after await

aliemjay:
## PR Type

Feature

## PR Checklist

## Overview

Based on #424

### DIFF
--- a/actix-server/src/server.rs
+++ b/actix-server/src/server.rs
@@ -8,7 +8,7 @@ use std::{
 
 use actix_rt::{time::sleep, System};
 use futures_core::{future::BoxFuture, Stream};
-use futures_util::stream::StreamExt;
+use futures_util::stream::StreamExt as _;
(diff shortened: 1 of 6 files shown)


actix-server/src/server.rs
```diff
@@ -228,67 +206,63 @@ impl ServerInner {
...
-                    .map(|worker| worker.stop(graceful))
-                    .collect::<Vec<_>>();
-
-                Some(Box::pin(async move {
-                    if graceful {
-                        // wait for all workers to shut down
-                        let _ = join_all(workers_stop).await;
-                    }
+                if graceful {
+                    // wait for all workers to shut down
```
robjtede, line 249: do we not want to send the stop messages even in the case of a force shutdown ?

aliemjay: oh, right!

robjtede (approved):

Issue #433, opened 2022-01-04
actix-tls v3 uses str_split_once and bumps MSRV to 1.52


david-mcgillicuddy-moixa:
Hi,
For reasons I don't really want to get into, for the time being we're stuck on rust version 1.51. Actix-tls v3 uses `str::split_once` (and actix-http uses `str::rsplit_once`) which bumps the MSRV up to 1.54. For the time being we've downgraded to actix-tls 3.0.0-beta.9 which doesn't do this, which is fine for now.

However, given that it's a small helper method, would you be open to a PR that replaces its use, or at least gates it behind some kind of compile flag? Happy to submit it. I realise we have just missed the beta/rc period too, awkward.
Otherwise I guess we'll have to maintain a fork for just those methods.

It is annoying that rust/cargo doesn't currently provide much feedback on this if you're not stuck on an old version yourself, although support is coming: https://github.com/rust-lang/rust/issues/65262

Thanks,
David

robjtede:
https://doc.rust-lang.org/std/primitive.str.html#method.split_once is marked as stable in 1.52. If you're really needing to stay on 1.51 then I'll consider a PR.

david-mcgillicuddy-moixa:
That's a good point thanks, I haven't tried a 1.52 upgrade (only 1.54 uncessfully) I'll do that now and get back to you.

I'm not trying to be mysterous or anything, the reason we're stuck on 1.51 is because we're stuck on Rocko version of Yocto build (for reasons I again don't especially want to get into) and meta-rust upgraded past rocko before upgrading to 1.54 (and straight past 1.52). We're working on upgrading our yocto build but given that this was some string helper fuctions it'll be a lot easier to switch out while the big build upgrade is ongoing.

If the 1.52 upgrade doesn't work I'll start on a PR thanks, would you prefer just removing the uses of `split_once` or feature flagging them so that users not stuck 6+ months in the past will see no changes after feature flag resolution?

david-mcgillicuddy-moixa:
[...]

PR #434, opened 2022-01-06
Replace str_split_once to lower actix-tls msrv to 1.50.0 and bump actix-net to 1.50.0

david-mcgillicuddy-moixa:
## PR Type

Bug Fix I think?

## PR Checklist

- [X] Tests for the changes have been added / updated.
- [X] Documentation comments have been added / updated.
- [X] A changelog entry has been made for the appropriate packages.
- [X] Format code with the latest stable rustfmt

## Overview

This PR replaces the use of `str::split_once`, which was stablised in 1.52.0 with `str::splitn`, keeping the MSRV at 1.50.0. This still isn't the 1.46.0 currently mentioned in the readme, due to the use of `bool_to_option` at `actix-server/src/server.rs:211` but I haven't gone that far back, I can do if preferred.

I could also gate this behind a conditional compilation directive if preferred so that users not stuck 6+ months in the past can use the new methods but also it's just a small string helper method so I don't believe that the additional complexity is worth it myself.

While this lowers the MSRV of `actix-tls` from 1.52.0 to 1.50.0 it raises the MSRV for the rest of the repo from 1.46.0 to 1.50.0 (due to no longer having a bunch of exclusions to the MSRV). I can revert that and keep everything on 1.46.0 if preferred. This way `actix-tls` doesn't have to be excluded from the MSRV requirement. Or I could just leave `actix-server` excluded and the rest of the repo can stay on 1.46.0 I guess.

Closes https://github.com/actix/actix-net/issues/433

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -23,4 +23,4 @@ ci-test-rt-linux = "    hack --feature-powerset test --package=actix-rt     --li
 ci-test-server-linux = "hack --feature-powerset test --package=actix-server --lib --tests --no-fail-fast -- --nocapture"
 
 # test lower msrv
-ci-test-lower-msrv = "hack --workspace --exclude=actix-server --feature-powerset test --lib --tests --no-fail-fast -- --nocapture"
+ci-test-lower-msrv = "hack --workspace --exclude=actix-server --exclude=actix-tls --feature-powerset test --lib --tests --no-fail-fast -- --nocapture"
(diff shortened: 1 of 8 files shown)


robjtede:
Can you put all the CI changes back please. I will accept this PR but don't want to change the official MSRV in CI even if it will build on the earlier compiler. That's to say I still recommend trying to update your toolchain.

david-mcgillicuddy-moixa:
Done now. And yes we very much still are it's just unfortunately a longer-term proposition.

robjtede (approved):

david-mcgillicuddy-moixa:
Great thanks

robjtede:
released this in actix-tls 3.0.1

PR #438, opened 2022-01-26
Replace `derive_more` with declarative macros

ibraheemdev:
## PR Type
Refactor

## PR Checklist

## Overview

Replace `derive_more` with declarative macros in actix-utils (to be used in -web).


robjtede:
Just in case something happens with the derive_more_lite idea, I inlined the macros into -tls to avoid adding things to the public interface of -utils.

PR #441, opened 2022-02-01
Add top level `block_on` helper

ibraheemdev:
## PR Type

Feature

## PR Checklist

## Overview

This is a useful helper, especially now that https://github.com/actix/actix-web/pull/2619 is merged.

### DIFF
--- a/actix-rt/src/lib.rs
+++ b/actix-rt/src/lib.rs
@@ -206,3 +206,20 @@ where
 {
     tokio::task::spawn_local(f)
 }
+
+/// Runs the provided future, blocking the current thread until the future completes.
+///
+/// This is a shortcut for creating a [`System`] and calling [`System::block_on`].
+///
+/// # Examples
+///
+/// ```
+/// fn main() {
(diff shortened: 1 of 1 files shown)


actix-rt/src/lib.rs
```diff
@@ -206,3 +206,20 @@ where
 {
     tokio::task::spawn_local(f)
 }
+
+/// Runs the provided future, blocking the current thread until the future completes.
+///
+/// This is a shortcut for creating a [`System`] and calling [`System::block_on`].
```
popzxc, line 212: I think it'll be helpful to add whether this method will panic or not if called within the scope of another `System`, similarly to [`tokio::Handle::block_on`](https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.block_on).

popzxc:
To give it another thought, does it really improve things much? Not so much verbosity is removed here, and I'm not sure whether it outweighs the cost of having one more item in the public API.

Issue #97, opened 2020-02-18
when backpressure happen, the uds file can not be rebuild


Right2Left:
https://github.com/actix/actix-net/blob/4f2910c6b301032d09bc2cedd0f259bb008788e1/actix-server/src/socket.rs#L137

Hi, in my project just found that, when server is restarting, but clients is still always retry connect.
In server, get in accept fuc before WorkerAvailability, so it check all worker is not available in accept_one then into backpressure, cleanup file path.
though WorkerAvailability soon, but client can not request anymore.

is that a bug or some configure need to set?

fakeshadow:
Sorry if I don't follow you. But is the desire behavior is to only clean the file when shutting down the server?

fakeshadow:
Since there is no further talk happen I would close this for now.

I think I have an idea of what the problem is but since I don't use uds much I'm not sure if it would fix the problem in the right way.

Please ping me for re-open if you have the same issue and want to help see it fixed.

PR #442, opened 2022-02-03
do not remove the uds socket on pause

### LINKED ISSUE #364: unix-fd listener should not remove sockets on deregister()
actix-server runs https://github.com/actix/actix-net/blob/605ec251432786b169490aac60679388aa902d95/actix-server/src/socket.rs#L74-L84

This `deregister` function can be called as part of flow control as response to `Pause` command https://github.com/actix/actix-net/blob/605ec251432786b169490aac60679388aa902d95/actix-server/src/accept.rs#L286-L292

When resuming operations, the socket file is gone and clients can not connect to the server anymore.

The cleanup should happen only when connection is closed finally (`Stop` command?)

hommeabeil:
## PR Type

Bug Fix

## PR Checklist

## Overview

Add a terminate function to the listener socket so the socket file is not remove in the deregister, but really when it is necessary. For a tcp socket, we have nothing to do and in the deregister, there is nothing left special to do for a UDS.

Closes #364 #97

### DIFF
--- a/actix-server/examples/tcp-echo.rs
+++ b/actix-server/examples/tcp-echo.rs
@@ -31,7 +31,7 @@ async fn run() -> io::Result<()> {
     let count = Arc::new(AtomicUsize::new(0));
 
     let addr = ("127.0.0.1", 8080);
-    info!("starting server on port: {}", &addr.0);
+    info!("starting server on: {}:{}", &addr.0, &addr.1);
 
     // Bind socket address and start worker(s). By default, the server uses the number of physical
     // CPU cores as the worker count. For this reason, the closure passed to bind needs to return
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -217,8 +217,6 @@ impl Accept {
                         self.deregister_all(sockets);
                     }
 
-                    self.terminate_all(sockets);
-
                     return true;
                 }
 
@@ -330,10 +328,6 @@ impl Accept {
             // Socket info with a timeout is already deregistered so skip them.
             .filter(|(timeout, _)| timeout.is_none())
             .for_each(|(_, info)| self.deregister_logged(info));
-    }
-
-    fn terminate_all(&self, sockets: &mut [ServerSocketInfo]) {
-        sockets.iter().for_each(|s| s.lst.terminate());
     }
 
     // Send connection to worker and handle error.
--- a/actix-server/src/accept.rs
+++ b/actix-server/src/accept.rs
@@ -217,6 +217,8 @@ impl Accept {
                         self.deregister_all(sockets);
                     }
 
+                    self.terminate_all(sockets);
+
                     return true;
                 }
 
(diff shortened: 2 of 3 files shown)


hommeabeil:
DO you think the CI error, might be related to [this](https://github.com/rust-lang/rust/issues/81890) ?

robjtede:
> DO you think the CI error, might be related

No idea, usually I ignore linker errors if they're not happening on all versions. It'll go away eventually.

fakeshadow:
You should use `Drop` trait. 

And my question in #97 still stands. The desired behaviour is still unclear to me. Does the file need to be cleared on server shutdown.

I'm asking this because the accept thread is notified to shutdown in non blocking manner and other part of the server would race to exit the program which mean the code you write may not get a chance to run at all.

hommeabeil:
The server should try to always clean is file when it is closing. This is what most server does and it leave the system in a cleaner state. However, we can't be ensure the socket file will be clean, because unlike a tcp socket, the kernel will not remove the file when sending a SIGKILL. So the server should handle this on startup (right now it does).

Does implementing the `Drop` trait will make the file remove blocking when stopping in a graceful manner (SIGTERM) ? I think this would be the desired behaviour.

fakeshadow:
Drop trait is the idiomatic way of clean up on resource go out of scope. Drop is not guaranteed to run either it's just more clear on the purpose and it's an implicit called compared to a homebrew method. So it's more about keep the code clean and readable.

As for block and wait for accept thread to exit you can look into this part of the code:
https://github.com/actix/actix-net/blob/72481313cc435aed95201ac1ff125fdf4d31f58b/actix-server/src/server.rs#L247

There are two possible ways to achieve a blocking cleanup:
1. Let `ServerInner` holds the JoinHandle of accept thread and block join when stop the server.
2. Make the `WakerInterest::Stop` notifier stateful where it can notify the `ServerEventMultiplexer` to proceed after the clean up and stop process is done.(For example using a channel)

fakeshadow:
https://github.com/actix/actix-net/pull/443

I made a PR for solution 1 in previous reply. If you don't mind you can rebase on it when it's accepted.

If you like to implement a different way in your PR it's also fine.

hommeabeil:
Thanks for your suggestion it is clear, I will look at your PR and look to rebase on it. I also re-think about if we should clean the file and I think their is some case where the server is not the owner of the file. One use case I have in mind is systemd socket activation, you receive a file descriptor not path, so you should let systemd manage the file.

Maybe we should clean the file only when the server receive a path, but if we receive a socket, we should not remove it since the server is not technically the owner... I will look at the different way to bind the server (I do not put too much attention the last time).

hommeabeil:
By looking around, I think we should leave the socket here. It would be cleaner to remove it on clean shutdown, but it will not be armful to leave it here. In fact this is what systemd does and even syslog. For a client this is not a problem because if you try to connect on a close socket, even if the file exist, you receive a Connection refused. If we need more we can always try to do the cleanup later, but for now, leaving the socket file will close the issue for everyone.

hommeabeil:
This is still affecting us, any news about this ? This prevent the use of systemd socket activation.

robjtede:
temporarily marking as draft due to conflicts

PR #463, opened 2022-07-15
normalize logs capital letter

leviska:
## PR Type

Refactor / Code Style

## PR Checklist

(not sure if my pr should tick other boxes, although some words in docs/code style would probably help)

## Overview

It seems that most packages in Rust uses lower case letters at the begging of the logs, and also Go code style recommends same style.
This crate uses a mix of upper and lower case, so I've decided to fix it to lower where appropriate. 
I've searched for `debug/trace/info/warn/error` keywords and changed the letters, if it isn't a common upper case abbreviation.

### DIFF
--- a/actix-server/src/server.rs
+++ b/actix-server/src/server.rs
@@ -193,14 +193,14 @@ impl ServerInner {
         let is_tokio = tokio::runtime::Handle::try_current().is_ok();
 
         match (is_actix, is_tokio) {
-            (true, _) => info!("Actix runtime found; starting in Actix runtime"),
-            (_, true) => info!("Tokio runtime found; starting in existing Tokio runtime"),
-            (_, false) => panic!("Actix or Tokio runtime not found; halting"),
+            (true, _) => info!("actix runtime found; starting in actix runtime"),
+            (_, true) => info!("tokio runtime found; starting in existing tokio runtime"),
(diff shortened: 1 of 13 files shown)


actix-server/src/server.rs
```diff
@@ -193,14 +193,14 @@ impl ServerInner {
         let is_tokio = tokio::runtime::Handle::try_current().is_ok();
 
         match (is_actix, is_tokio) {
-            (true, _) => info!("Actix runtime found; starting in Actix runtime"),
-            (_, true) => info!("Tokio runtime found; starting in existing Tokio runtime"),
+            (true, _) => info!("actix runtime found; starting in Actix runtime"),
+            (_, true) => info!("tokio runtime found; starting in existing Tokio runtime"),
```
JohnTitor, line 197: Capitalization here is inconsistent, mixed use of actix/Actix and tokio/Tokio.

JohnTitor:
criterion's dependency started to use edition 2021, I think testing on MSRV has some benefits (e.g. making sure it works by tests), and bumping MSRV to 1.56 would be fine (actix-web's MSRV is 1.57, so it shouldn't make much difference IMO). @robjtede Any objections?

actix-server/src/server.rs
leviska: fixed

robjtede (approved):

PR #466, opened 2022-09-10
add MPTCP socket protocol (optional)

Martichou:
## PR Type
Feature

## Overview
Add the possibility to use the MPTCP protocol at the socket level for users of ServerBuilder.

MPTCP is now more widely available since Linux Kernel version >= 5.6. But it still need to be enabled manually using: `sysctl net.mptcp.enabled=1` (of course, MPTCP is only available on Linux).

The new MPTCP struct give the user the option to determine how we'll handle the case where MPTCP is not available on the host, either we crash, or we fallback to regular TCP.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ jobs:
           - { name: Windows (MinGW), os: windows-latest, triple: x86_64-pc-windows-gnu }
           - { name: Windows (32-bit), os: windows-latest, triple: i686-pc-windows-msvc }
         version:
-          - 1.60.0
+          - 1.63.0
           - stable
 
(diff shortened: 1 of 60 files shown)


Martichou:
_note: this PR will fail until socket2 make a new release with the new Protocol::MPTCP, but the content can still be reviewed_

Martichou:
Unfortunately, the `socket2` version needed is the `0.5`.

I see there was a comment `# TODO(MSRV 1.64) update to 0.5`. What is the status on that? (cc @robjtede)
This is the only thing still needed for MPTCP in Actix :1st_place_medal:

robjtede:
Thanks for pushing this. My tweaks are mostly mechanical. We're expecting Windows CI failures atm so no worries there.

PR #467, opened 2022-10-10
remove hard dependency for tokio net and tokio signal

### LINKED ISSUE #541: Conditional inclusion of tokio-net and tokio-signal for targets that do not support them.
## Expected Behavior

Actix should be able to compile on system that do not support tokio-net and tokio-signal.

## Current Behavior

tokio net and signal features are currently hard-included from the actix-rt crate, although actix-rt barely uses them.

## Possible Solution

Create a feature gate that will enable inclusion of tokio-signal and tokio-net. Make it included by default, to remain backward compatible.

tokio-signal is only pub used from actix-rt::signal, and that module should only be compiled if tokio-signal is explicitly asked for by the user of the actix library.

actix::net provides the ActixStream, and could be left out, if unused, although I'm not particularly sure why tokio::net is required for ActixStream support. The same gate should be used for default_tokio_runtime calling enable_io or not.

## Steps to Reproduce (for bugs)

include actix = "*" for a "riscv32imc-esp-espidf" target, using esp-idf OS, using std.

Can try to provide a minimal repro project if necessary, although I don't think there is need for one.

## Your Environment

Compiling rust for an ESP32C3 target (RISCV).

* Rust Version: rustc 1.65.0-nightly (750bd1a7f 2022-09-14)
* Actix Version: 0.13.0

axos88:
Closes https://github.com/actix/actix/issues/541.

Change is backward compatible, as the new net feature is enabled by default


robjtede:
temporarily marking as draft due to CI failure

robjtede:
superseded by https://github.com/actix/actix-net/pull/973

PR #471, opened 2022-11-10
fix no_std compatibility

VladimirBramstedt:
The bytes crate is only no-std with no default features... if you try to use bytestring for an actual no_std target, it wont work. 
Bug fix: make sure we depend on bytes with no default features.
suggestion: Add step in CI that tries to build for a no_std target, to make sure no_std compatibility is kept in the future.


JohnTitor (approved):
Thank you.

PR #484, opened 2023-05-31
Add getter methods for actix_rt::Runtime and tokio::runtime::Runtime

In-line:
## PR Type

Feature

## PR Checklist

## Overview

The `tokio_runtime()` method retrieves a reference to the underlying Tokio runtime associated with an `actix_rt::Runtime` instance. This allows direct interaction with Tokio's features. It is useful for sharing the same runtime between different modules of a project, enabling tasks to be spawned on the same runtime. The method returns an immutable reference to the `tokio::runtime::Runtime`.

Also, as a follow-up feature, `runtime()` is added on `actix_rt::SystemRunner`.


robjtede (approved):

robjtede:
Don't see a reason to deny adding these methods. Thanks for the contribution and especially the well written docs!

PR #480, opened 2023-04-01
Rustls v0.21 support

robjtede:
## PR Type

Feature

## PR Checklist
Check your PR fulfills the following:

- [x] Tests for the changes have been added / updated.
- [x] Documentation comments have been added / updated.
- [x] A changelog entry has been made for the appropriate packages.
- [x] Format code with the latest stable rustfmt

## Overview

Rustls v0.21 support.

PR #487, opened 2023-08-22
Update dependencies for RUSTSEC-2023-0052

AldaronLau:
## PR Type

[Advisory](https://rustsec.org/advisories/RUSTSEC-2023-0052)

## PR Checklist

## Overview

Due to the fact that it is a security vulnerability, this updates public dependencies (therefore includes breaking changes).  Behavior is mostly the same, except that the error message now calls out that it's actix-tls that doesn't support non-hostnames, rather than rustls (now that that has been fixed).  I also changed the `ErrorKind` to `InvalidInput`, as I believe that to be more accurate.

I understand that it may be desirable for a number of reasons to update and finish https://github.com/actix/actix-net/pull/480 instead of merging this PR (feel free to close), but my personal opinion is that 4.0 is warranted.

### DIFF
--- a/actix-tls/CHANGES.md
+++ b/actix-tls/CHANGES.md
@@ -3,6 +3,8 @@
 ## Unreleased - 2023-xx-xx
 
 - Minimum supported Rust version (MSRV) is now 1.65.
+- Update tokio-rustls to 0.24
+- Update webpki-roots to 0.25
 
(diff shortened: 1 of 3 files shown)


robjtede:
doing this dependency update in a breaking manner isn't an acceptable way forward for this crate; the existing work in #480 has now been updated

PR #573, opened 2024-06-12
Fix logging "starting service:..."

oliverjantar:
## PR Type

Bug Fix 

## PR Checklist

## Overview
When starting a server, message `"starting service: "{}", workers: {}, listening on: {}"` is never logged. 
The issue happens when [mapping](https://github.com/actix/actix-net/blob/912daa3d0aeb2bc09889a6d200ccef839ad6764a/actix-server/src/server.rs#L186) a `builder.sockets` and using `mem::take()` that replaces the original value with default. Since `builder.sockets` is empty, it won't log anything in the [for cycle](https://github.com/actix/actix-net/blob/912daa3d0aeb2bc09889a6d200ccef839ad6764a/actix-server/src/server.rs#L201).

### DIFF
--- a/actix-server/src/server.rs
+++ b/actix-server/src/server.rs
@@ -183,11 +183,6 @@ impl ServerInner {
     }
 
     fn run_sync(mut builder: ServerBuilder) -> io::Result<(Self, ServerEventMultiplexer)> {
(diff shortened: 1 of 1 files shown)


robjtede (approved):
well spotted, thanks 👍🏻

PR #599, opened 2024-09-04
Change with_tokio_rt to accept Arc<Runtime>.

gz:
This allows to share tokio runtimes across different sub-systems inside your application.

## PR Type

Feature, Refactor

https://github.com/actix/actix-net/issues/580

## PR Checklist

## Overview

The current signature of https://docs.rs/actix-rt/latest/actix_rt/struct.System.html#method.with_tokio_rt expects a `Runtime`. It would be better if I can supply an Arc<Runtime> instead. The reason being that is that in a bigger application we have multiple systems trying to leverage tokio. With the given API I still need to create a separate runtime for actix. Having the same runtime is helpful for better control over CPU resources as there aren't `n` runtimes competing for CPU time.

FWIW this is a breaking change so if there alternative ways to achieve this I'm happy to give those a try too.

### DIFF
--- a/actix-rt/examples/multi_thread_system.rs
+++ b/actix-rt/examples/multi_thread_system.rs
@@ -10,6 +10,7 @@ fn main() {
             .worker_threads(2)
             .enable_all()
             .build()
+            .map(std::sync::Arc::new)
             .unwrap()
     })
     .block_on(async_main());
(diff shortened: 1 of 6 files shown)


robjtede:
To avoid breaks we could add a new method for this.

robjtede:
closing due to inactivity and that actix-rt will likely be deprecated soon

PR #852, opened 2026-04-11
perf(bytestring): improve `From<ByteString> for String` performance

JohnTitor:
## PR Type

INSERT_PR_TYPE

## PR Checklist

- [ ] Format code with the latest stable rustfmt

## Overview

`ToString` requires an extra allocation when converting: https://github.com/rust-lang/rust/blob/c7c14d4fb0310429e05e1a24dcf879053ee25b09/library/alloc/src/string.rs#L2911

Since we assume the value is UTF-8, it should be okay to use `String::from_utf8` just like other convension.

Verified the effect roughly with:
```rs
const STRING_LEN: usize = 65536;

fn owned_bytestring() -> ByteString {
    ByteString::from("a".repeat(STRING_LEN))
}

fn bench_string_conversion(c: &mut Criterion) {
    c.bench_function("bytestring_into_string_from_owned_buffer", |b| {
        b.iter_batched(
            owned_bytestring,
            |value| black_box(String::from(value)),
            BatchSize::SmallInput,
        );
    });
}
```

### DIFF
--- a/bytestring/src/lib.rs
+++ b/bytestring/src/lib.rs
@@ -6,11 +6,7 @@
 
 extern crate alloc;
 
-use alloc::{
-    boxed::Box,
-    string::{String, ToString},
-    vec::Vec,
-};
+use alloc::{boxed::Box, string::String, vec::Vec};
 use core::{borrow::Borrow, fmt, hash, ops, str};
 
 use bytes::Bytes;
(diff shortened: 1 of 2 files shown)


bytestring/src/lib.rs
```diff
@@ -189,7 +185,7 @@ impl From<Box<str>> for ByteString {
 impl From<ByteString> for String {
     #[inline]
     fn from(value: ByteString) -> Self {
-        value.to_string()
+        String::from_utf8(value.0.into()).expect("ByteString invariant violated")
```
robjtede, line 188: given we impl `Deref` using unsafe, it wouldn't be crazy to use `String::from_utf8_unchecked` here

JohnTitor: Ah great idea, will do that.

PR #865, opened 2026-04-26
perf(bytestring): micro-improve performance against ASCII conv

JohnTitor:
## PR Type

INSERT_PR_TYPE

## PR Checklist

- [ ] Format code with the latest stable rustfmt

## Overview

Another one, improve performance against ASCII to avoid UTF-8 convension. On local benchmarks, it shows roughly 20-30% improvement.
Since I assume most values would be ASCII, I expect this would be benefitical for most usecases.


robjtede (approved):
cool, TIL about slice::is_ascii

PR #870, opened 2026-04-30
perf(codec): improve performance by storing next index

JohnTitor:
## PR Type

perf

## PR Checklist

## Overview

Improve performance from `O(n^2)` to `O(n)` by storing next_index in `LinesCodec`. The idea is from https://docs.rs/tokio-util/latest/src/tokio_util/codec/lines_codec.rs.html.
Didn't touch the changelog as it's unreleased yet.


robjtede:
thoughts on a `.reset()` method, in case someone legit wants to use the same struct for a new buffer ?

JohnTitor:
After some thinking of this, it'd be a bit tricky to support such a usecase as it might cause index corruption. In https://github.com/actix/actix-net/pull/870/changes/d26ccb6c0bdf0b054e95c3aa18737673d5cf9a25, I tightened its availability for external use for now. Maybe we could revisit this topic if someone wants to use it.

robjtede (approved):

PR #930, opened 2026-08-03
Propagate graceful shutdown to server services

robjtede:
## Summary

- add a sticky, read-only `GracefulShutdownSignal`
- expose it through `ServerBuilder::graceful_shutdown_signal()`
- notify listeners before workers begin a graceful shutdown

## Why

Connection service futures are detached from the worker shutdown command. Protocol dispatchers therefore need a server-owned notification to stop accepting work and drain their current operation while the worker waits for active connections.

The signal is emitted for every graceful stop source, including OS signals, `ServerHandle::stop(true)`, and `ServerBuilder::shutdown_signal()`.

## Validation

- `cargo test -p actix-server --lib -- --nocapture`
- `just clippy`
- `cargo test -p actix-server --doc`


JohnTitor (approved):
Looks good!

Issue #597, opened 2024-09-03
Use the new tokio `LocalRuntime` in `actix_rt`


Sytten:
**This is a tracking issue**

Tokio is deprecating the `LocalSet` (https://github.com/tokio-rs/tokio/issues/6741) in favour of a new `LocalRuntime` (https://github.com/tokio-rs/tokio/issues/6739) which will avoid panics if a user mixes `tokio::spawn` with `tokio::spawn_local`/`actix_rt::spawn`. It should also improve performances.

Once this is released we should migrate to the new runtime.

Noah-Kennedy:
This is also likely to improve your performance a bit.

robjtede:
It's also possible that LocalRuntime will be easy enough to use that we can just make actix-rt optional altogether.

Noah-Kennedy:
> It's also possible that `LocalRuntime` will be easy enough to use that we can just make actix-rt optional altogether. In particular, if there's a `#[tokio::main(flavor = "local")]` or similar.

You should leave this in the LocalSet issue thread as well so folks know that the macros are something you care about.

Sytten:
We will still need the system and arbiter concepts so pretty sure we will keep the crate around for some time.

robjtede:
We'd only need them for actors. The crate can stay around but it would be removed from the dependency tree of `actix-web` in a major release.

Darksonn:
Is this something we can do now? I looked a bit at the actix-rt crate, and the API is a bit difficult to work with as it has constructors that take values of type `Runtime` rather than the new `LocalRuntime` type.

Sruhvx-jpg:
u could patch this post gta 6's release so we could say "we got gta 6 before actix net's single patch"

robjtede:
Instead of breaking actix-rt, we will likely deprecate it soon and it won't be used in Actix Web v5.

`actix-server` (when running in pure Tokio mode) now uses `LocalRuntime` for workers.

Sytten:
So no Arbiter anymore I am guessing? Aka actix actor wont run on those.

robjtede:
`actix` is kinda soft-deprecated. In -web v5 planning I'd been thinking to drop -rt altogether in which case both crates would be marked deprecated officially.

I don't personally want to maintain the actor system any more.

Sytten:
I need to migrate off of it myself so this pushes it in the right direction.

PR #943, opened 2026-08-24
feat: use Tokio LocalRuntime for server workers

robjtede:
## Summary

- use Tokio `LocalRuntime` for worker initialization and execution in a plain Tokio runtime
- remove cross-thread `Handle::block_on` and `LocalSet` from that worker path
- raise the Tokio minimum to 1.51 and cover timer-backed factory initialization

Refs #597

### DIFF
--- a/actix-server/src/worker.rs
+++ b/actix-server/src/worker.rs
@@ -18,9 +18,12 @@ use actix_rt::{
     Arbiter, ArbiterHandle, System,
 };
 use futures_core::{future::LocalBoxFuture, ready};
-use tokio::sync::{
-    mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
-    oneshot,
+use tokio::{
+    runtime::{Builder, LocalOptions},
+    sync::{
+        mpsc::{unbounded_channel, UnboundedReceiver, UnboundedSender},
+        oneshot,
+    },
 };
 use tracing::{error, info, trace};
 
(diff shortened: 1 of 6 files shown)


robjtede:
@Darksonn is it this easy? Anything I should add to the changelog w.r.t. migrations or possible compat issues?

actix-server/src/worker.rs
```diff
@@ -305,17 +308,20 @@ impl ServerWorker {
...
+                        let rt = Builder::new_current_thread()
+                            .enable_all()
+                            .max_blocking_threads(config.max_blocking_threads)
+                            .build_local(LocalOptions::default())
+                            .unwrap();
 
-                        // init services using existing Tokio runtime (so probably on main thread)
-                        let services = rt_handle.block_on(ls.run_until(async {
+                        // init services using the worker's local runtime
+                        let services = rt.block_on(async {
```
Darksonn, line 324: It looks like this changes the behavior slightly by introducing a per worker `spawn_blocking()` pool, whereas previously it used the pool from the existing runtime.

Darksonn (commented):
Other than one thing, this seems like it shouldn't change anything else.

actix-server/src/worker.rs
robjtede: Thanks for looking. I'll have a think about the observable effects of that.

Darksonn: Wait ... the old code also creates a new runtime. So you're only changing what happens during the `factory.create()` call. Once you get past that, the old code switches to using a separate runtime too, so only the factory creation is even affected to begin with.

robjtede: Yea that's what I'm thinking. Could actually be seen as a bug in existing code. I can imagine factory create stage doing CPU pinning for example.

PR #953, opened 2026-09-03
fix(actix-server): clean up acceptor and worker threads on drop

### LINKED ISSUE #3765: Server never releases open ports
Thank you for the awesome work of actix-web authors!

## Context

Trying to create a reliable application, I was using the strategy to react to unrecoverable errors by reinitializing the whole application and retry the workflow preventing a whole shutdown and process restart. The implementation went just fine until I had to add a webserver as an optional feature. No matter how I tried to shut down the HttpServer, the bound ports were never released, preventing the application from properly reinitializing.

The only related discussion I've found was in https://github.com/actix/actix-web/discussions/3025 .

## Expected Behavior

There should be a way to properly close the server and release the Http(s) ports bound during server initialization.
Consequently, server instances initialized later should be able to successfully bind it again.

## Current Behavior

No matter how I tried closing the server by dropping the server instance, shutting it down with a `ServerHandle.stop()`, etc, the port was always kept bound. 

## Possible Solution

Maybe there is already a proper way to unbind the port, but I haven't found it anywhere in the docs as well. If it's not present yet, I think it should be added and documented.

## Your Environment

I am using rustc 1.86 with actix-web 4.11.

Sruhvx-jpg:
Dropping or cancelling a `Server` future detached the background acceptor OS thread without stopping it, leaving listening TCP sockets open and causing subsequent binds to fail with `EADDRINUSE`.

Huge thanks to **Istvan Zolyomi** (@izolyomi) for isolating and reporting this bug in https://github.com/actix/actix-web/issues/3765, where dropped/cancelled servers left ports permanently bound in resilient reinitialization workflows.

### Summary of Changes
- Implements `Drop for ServerInner` to wake the acceptor's `mio::Poll` loop, stop workers, and join the thread so socket file descriptors are closed and the port is freed immediately.
- Adds an integration test ensuring cancelled server tasks release ports immediately.
- Adds changelog entry in `actix-server/CHANGES.md` (crediting @izolyomi).

Fixes https://github.com/actix/actix-web/issues/3765
Refs https://github.com/actix/actix-web/discussions/3025
Reported-by: @izolyomi

### DIFF
--- a/actix-server/CHANGES.md
+++ b/actix-server/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Clean up acceptor and worker threads when `Server` is dropped or cancelled, immediately releasing bound listening sockets and ports. Reported by Istvan Zolyomi ([@izolyomi](https://github.com/izolyomi)).
+
 ## 2.9.1
 
 - Fix race condition where accept thread shutdown cause cause workers to be dropped prematurely.
--- a/actix-server/src/server.rs
+++ b/actix-server/src/server.rs
@@ -386,3 +386,21 @@ impl Stream for ServerEventMultiplexer {
         this.cmd_rx.poll_recv(cx)
     }
 }
+
+impl Drop for ServerInner {
+    fn drop(&mut self) {
+        if !self.stopping {
+            self.stopping = true;
+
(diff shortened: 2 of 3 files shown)


robjtede:
superseded by #966 

thanks for the contribution

PR #966, opened 2026-09-10
fix(actix-server): clean up cancelled servers during shutdown

### LINKED ISSUE #3765: Server never releases open ports

robjtede:
## Overview

Dropping or cancelling a running `Server` previously detached its acceptor thread and left listening ports bound. Close the listeners before drop returns and request forced worker shutdown, including when cancellation interrupts an unfinished graceful shutdown.

Use the outstanding acceptor join handle to detect incomplete cleanup. If shutdown has already started, force-stop workers and join the acceptor without waking its potentially closed poll instance again. Normal explicit non-graceful shutdown remains unchanged. Document that drop waits for the acceptor but does not wait for worker threads to exit.

Supersedes and obsoletes #953. That proposal covers cancellation while running, but its `!stopping` guard skips cleanup when cancellation interrupts graceful shutdown. Thanks to @Sruhvx-jpg for the original proposal and @izolyomi for reporting the issue.

Fixes https://github.com/actix/actix-web/issues/3765

## PR Type

Bug Fix

## PR Checklist

- [x] Format code with rustfmt (repository nightly recipe).

## Validation

- `cargo test --offline -p actix-server`: 28 passed, 1 ignored on macOS.
- `just all_crate_features='' clippy`: workspace and all targets, default features.
- `cargo +nightly fmt -- --check`.
- The listener-release regression failed with `AddrInUse` before cleanup was added. The graceful-shutdown cancellation regression failed with the original proposal's guard and passes with this implementation.
- Tests cover dropping a running server, cancelling graceful shutdown with an active connection, and explicit non-graceful shutdown. Both shutdown tests verify connection closure and port reuse.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,6 +57,7 @@ dependencies = [
  "socket2",
  "static_assertions",
  "tokio",
+ "tokio-test",
  "tokio-util",
  "tracing",
  "tracing-subscriber",
@@ -2436,6 +2437,7 @@ dependencies = [
  "futures-sink",
  "libc",
  "pin-project-lite",
+ "slab",
  "tokio",
 ]
 
--- a/actix-server/CHANGES.md
+++ b/actix-server/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- Close listeners and request forced worker shutdown when a running server is dropped, including during graceful shutdown.
+
 ## 2.9.1
 
 - Fix race condition where accept thread shutdown could cause workers to be dropped prematurely.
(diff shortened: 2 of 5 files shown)


robjtede:
I ran the actix-web test suite against this patch and all passed.

PR #971, opened 2026-09-12
feat(service): implement Transform for Option

### LINKED ISSUE #465: impl Transform for Option
This topic continues https://github.com/actix/actix-web/pull/2858 , whole goal is to `App::wrap` an optional middleware. `Condition` middleware exists but requires a transformer even when unused. It seems to me that using an `Option` was more natural. But as that PR was on a different crate than `Transform`'s, a newtype was introduced, and `.into()` required an explicit annotation. Sadly this was non-optimal.

@fakeshadow suggested that this feature should rather be moved into `Transform`'s crate, so that `.wrap` can natively take an `Option<SomeMiddleware>`. This would mean moving the `ConditionMiddleware` code too.

robjtede:
## PR Type

Feature

## PR Checklist

## Overview

Implement `Transform<S, Req>` for `Option<T>`. `Some` applies the middleware, while `None` uses the inner service directly. This supports optional middleware without a newtype or a separate conversion. The middleware must preserve the inner service's response and error types.

The implementation uses allocation-free service and future enums and retains `no_std` support. Unit tests cover enabled and disabled middleware, lazy construction, initialization errors, readiness, call errors, and pending futures.

Closes #465.


actix-service/CHANGES.md
```diff
@@ -2,8 +2,9 @@
 
 ## Unreleased
 
+- Implement `Transform` for `Option<T>` to support optional middleware that preserves response and error types.
+- Use `core::future::{ready, Ready}` instead of equivalent from `actix-utils`.
```
robjtede, line 6: @JohnTitor I'm extremely tempted to risk this tiny breakage on the assumption that no one is writing:

```rust
struct Foo { fut: actix_utils::future::Ready };

let a: FnServiceFactory<F, Fut, Req, Res, Err, Cfg> as ServiceFactory<Req>::Future = ...;

Foo {
  fut: a
}
```

(or whatever... I can't even contrive of a setup that would ever get written)

If anyone files issue we can do an immediate yank and probably revert to the actix-utils one.

WDYT?

PR #976, opened 2026-09-13
test(tls): verify handshake timeout on stalled connection in rustls

Sruhvx-jpg:
## PR Type
Tests

## Overview
Add a regression test verifying that `Acceptor` enforces `set_handshake_timeout` by returning `TlsError::Timeout` when an incoming connection stalls before completing the TLS handshake.

Also decouples `tests/accept-rustls.rs` from the unnecessary file-level `feature = "openssl"` requirement so pure-Rustls tests can run without C OpenSSL dev dependencies.


robjtede (approved):
lgtm 👍🏻

PR #977, opened 2026-09-13
test(tls): verify rejection of untrusted client certificate in rustls

Sruhvx-jpg:
## PR Type
Tests

## Overview
Add an end-to-end integration test verifying that `Acceptor` configured with client certificate verification (`WebPkiClientVerifier`) properly rejects untrusted client certificates during the TLS handshake, cleanly emitting `TlsError::Tls` with `io::ErrorKind::InvalidData`.

Also decouples `tests/accept-rustls.rs` from the unnecessary file-level `feature = "openssl"` requirement so Rustls integration tests can run independently.

*Note: The integration test case and the client verifier helper struct were written by AI. If you think this is slop, just reject it.*


robjtede:
too much code to add coverage to only 1 addition line

PR #978, opened 2026-09-13
test(tls): verify rejection of plaintext and invalid record framing in rustls

Sruhvx-jpg:
## PR Type
Tests

## Overview
Add integration tests verifying that `Acceptor` properly rejects incoming connections with non-TLS wire payloads:
1. Plaintext HTTP requests (`GET / HTTP/1.1`) where the first record byte is `'G'` (`0x47`).
2. Corrupted record headers with invalid content type bytes (blank space padding `0x20`).

Both scenarios assert that `Acceptor` bubbles up `TlsError::Tls(io_err)` with `io::ErrorKind::InvalidData`.

Also decouples `tests/accept-rustls.rs` from the unnecessary file-level `feature = "openssl"` requirement so Rustls tests can run independently.


robjtede:
doesn't add test coverage to any new lines after #982 merge

PR #979, opened 2026-09-13
perf(codec): allocate framed buffers on demand

robjtede:
`Framed::new` allocates two 8 KiB buffers before the transport is polled. This change starts both buffers empty and lets the existing read/write reserve paths allocate storage when needed. An unused transport now requests zero heap storage for buffers, and a read-only or write-only transport avoids the unused direction's 8 KiB allocation.

The first read/write reserve policy and the 8 KiB write-readiness threshold are unchanged. Active bidirectional transports still allocate both buffers. This is primarily an idle/one-sided memory improvement, not a claim of faster network I/O.

## Changes

- Initialize both buffers with `BytesMut::new()` and document allocation on demand.
- Add regression tests for unused, read-only, and write-only transports. These three assertions fail on the original constructor and pass after the change.
- Test wakeups after pending reads and 72 request/response exchanges across 18 combinations of message size and duplex capacity, including a one-byte I/O buffer and 64 KiB messages. Verify contents and EOF after shutdown.
- Add a focused Divan allocation benchmark and a Criterion benchmark comparing the default constructor with explicit 8 KiB buffers through `FramedParts`.
- Enable Tokio runtime, macros, network, and timer features for tests/benchmarks only.

## Measurements

macOS ARM64, rustc 1.99.0-beta.4, workspace release profile (opt-level 3, LTO, one codegen unit). Both constructor policies run in the same uninstrumented binary. The final constructor/duplex confirmation uses 20 samples, 0.3 s warmup, and 1 s target measurement; the longest duplex cases extend to about 8.6 s to collect 20 samples. Warm TCP uses 30 samples, 0.5 s warmup, and 1.5 s measurement. Reported values are point estimates, not request percentiles.

| Workload | Explicit 8 KiB buffers | Lazy/default |
| --- | ---: | ---: |
| Construct and drop unused transport | 37.05 ns | 3.48 ns |
| Cold duplex: one 64 B request/response | 66.68 µs | 66.45 µs |
| Duplex: 64 × 64 B request/responses | 5.168 ms | 5.040 ms |
| Cold duplex: one 8 KiB request/response | 6.690 ms | 6.697 ms |
| Duplex: 64 × 8 KiB request/responses | 427.67 ms | 434.36 ms |
| Warm TCP: one 64 B request/response | 16.23 µs | 16.51 µs |
| Warm TCP: 64 × 64 B request/responses | 1.048 ms | 1.051 ms |

Duplex capacity is deliberately only 64 bytes, so the large-message cases spend most of their time handling backpressure. TCP uses established loopback connections and TCP_NODELAY; connection establishment is outside timing. The initial per-iteration TCP-connect fixture was discarded because port reuse stalls made it unsuitable for sustained benchmarking. The retained TCP fixture avoids that problem.

The active-I/O point estimates are within about 2.5% of one another. Earlier runs changed the direction of some differences, so these results do not establish an I/O speed improvement. The deterministic benefit is removal of two allocations and 16,384 requested bytes before first use. Divan is measured separately because its allocator instrumentation affects timing; requested bytes exclude allocator metadata, kernel buffers, and RSS.

## Validation

- macOS ARM64: crate tests on the default compiler and Rust 1.88.0 (MSRV).
- Linux ARM64: crate tests in `rust:1.88-bookworm`, with the source mounted read-only.
- All 17 actix-codec tests pass. The new initialization, wakeup, partial-I/O, content, and shutdown checks pass.
- Full workspace `just test`: 128 passed with no default features; 149 passed with all features; one skipped test in each pass.
- Workspace `just clippy`, MSRV compilation of all targets, nightly Rust formatting, and TOML formatting pass.

Reproduction:

```sh
cargo test -p actix-codec
cargo +1.88.0 test -p actix-codec
cargo bench -p actix-codec --bench framed_allocations -- --sample-count 10 --sample-size 1
cargo bench -p actix-codec --bench framed -- --sample-size 30 --warm-up-time 0.5 --measurement-time 1.5
```

This PR changes no codec behavior or public type signatures. Buffer capacities are observable through `into_parts`; callers now see empty initial buffers. Callers that explicitly need preallocated storage can continue to supply it through `FramedParts`. Windows and other CPU architectures are left to CI; no cross-platform timing claim is made.

First use now pays the allocation cost previously paid by construction. The `framed first use` cases isolate that trade-off by excluding constructor and output-drop time with Criterion batching. It is not a claim that first-read or first-write latency falls merely because construction is cheaper.

| Isolated first use, constructor/drop excluded | Preallocated | Lazy/default |
| --- | ---: | ---: |
| Write `hello` | 7.80 ns | 18.38 ns |
| Read `hello\n` | 45.29 ns | 58.29 ns |

These cases use `BatchSize::LargeInput`, 30 samples, 0.5 s warmup, and 1.5 s measurement. First-use allocation adds about 11 ns to this write and 13 ns to this read on the measured host, in exchange for avoiding up-front allocations for unused directions. The final all-target Clippy check also passes with `-D warnings`.


robjtede:
bad change given we don't assume `Framed` is thrown away before writing into buffers, only potential change is adding a `with_capacity` so callers can decide, but i don't think its worth it without a feature request

